Issue #5, opened 2022-09-26
Requirements for Interoperability of Services using User Identity


frederikrothenberger:
The working group is currently in the process of gathering requirements for the [interoperability topic described here](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/interoperability.md#interoperability-of-services-using-user-identity
).

To start the discussion about requirements I'd like to propose the following set of requirements:

| # | Requirement | Description | Source |
| -- | -- | -- | -- |
R1 | Dapps can specify permission sets that can be requested by other (third party) services. | The permission sets should consist of the following information:<br>* Metadata<br>** Human readable title and description.<br>*** Developers should be able to provide this in multiple languages.<br>*** The description should be able to reference requested values (i.e. “Allow access to account **1234**?”).<br>** Identifier to refer to the given permissions.<br>* Restrictions<br>** Restrictions on canister calls<br>*** Allowed functions<br>*** Allowed arguments<br>**** (multiple) value ranges<br>**** Set of values<br>**** Single values<br>**** Patterns?<br>** Restrictions on validity<br>*** Max allowed validity period |  
R2 | Permissioned delegations allow a service to access another using the application-specific principal of the other service. | Given a user that has delegated some permissions from service B to service A.<br>Service A will be able to:<br>* access service B using the application-specific principal of service B for actions allowed within the limits set by the granted permissions.<br>* access service B using the application-specific principal of service A with unrestricted permissions (current situation). |  
R3 | Permissioned delegations should reveal as little information as possible about the attributes of the user. | A service should learn as little information as possible from the permissioned delegation about the user whose permission it received.<br>(However, this does not prevent services from revealing the user principals themselves.) |  
R4 | Once granted, a dapp should not have to request the same permissions again within the validity period of the last delegation. | This requires IDPs to keep track of issued delegations. Example:<br>1. Service A authenticates with IDP → delegation for principal A<br>2. Service A requests indefinite access to some permissions of Service B<br>2.a) User grants consent to allow A to access B given the restrictions.<br>2.b) A does not learn the dapp specific principal of the User on Service B (see R2).<br>3. User logs out of service A.<br>4. User logs into service A again.<br>4.a) The IDP will automatically issue a combined delegation:<br>4.a)i Delegation for principal A.<br>4.a)iI Delegation for the dapp specific principal of Service B restricted to the permissions consented to in step 2. |  
[...]

PR #6, opened 2022-09-27
Add requirements to interoperability document

frederikrothenberger:
This PR aims to add concrete requirements to the interoperability discussion. It will only be merged once all concerns raised are properly addressed and the working group has formed consent on the given set of requirements.

This is PR has been created from the set proposed in #5 but already incorporated the feedback there.

Rendered table for easier viewing: https://github.com/dfinity/wg-identity-authentication/blob/frederik/req-interop/topics/interoperability.md#requirements


michelabdalla-dfinity (approved):
Looks good to me.

PR #11, opened 2022-12-01
Attribute sharing - Added details and examples for the message flows

michelabdalla-dfinity:


frederikrothenberger (approved):
Thanks a lot!

PR #32, opened 2023-07-14
airgap/dapp-wallet-messages

jsamol:
This PR proposes a message standard that defines communication between dApps and wallets as described in [RFP-7](https://forum.dfinity.org/t/assigned-rfp-7-wallet-standard-reference-implementation/20024).


jsamol:
The initial commit defines possible messages focusing on the content.

As already discussed in the last working group meeting, there are some concerns regarding the `permission` messages and whether they are needed in the IC ecosystem at all. This, however, requires further discussion as the initial conclusion was that such messages are redundant in the light of Internet Identity broad use, but some possible applications could be found.

jsamol:
I've looked deeper into Internet Identity, and I'm not sure if it can be used to authorize requests to wallets as defined in RFP-7 🤔 

Correct me if I'm wrong, but it seems like Internet Identity is the only source of its accounts, meaning a user can either create a new account/identity or import one that was previously created with Internet Identity. I can't see any options to import an account created in a wallet or loosely connect a wallet account to Internet Identity without revealing the passphrase used to create the account. This means that the only connection between Internet Identity and a wallet could be created only if the wallet imported the account from Internet Identity, which I don't think is possible in every case.

I believe that what we're defining here, through this wallet standard, is a secondary method of authentication in IC, an alternative to Internet Identity, where the source of the account can be any wallet that implements the standard. If that's the case, then the `permission` messages are indispensable as their main role is to obtain account related data from the wallet.

It would be great, though, to hear from someone with a broader knowledge of the whole IC ecosystem to validate my thoughts 😅

frederikrothenberger:
> I've looked deeper into Internet Identity, and I'm not sure if it can be used to authorize requests to wallets as defined in RFP-7 🤔
> 
> Correct me if I'm wrong, but it seems like Internet Identity is the only source of its accounts, meaning a user can either create a new account/identity or import one that was previously created with Internet Identity. I can't see any options to import an account created in a wallet or loosely connect a wallet account to Internet Identity without revealing the passphrase used to create the account. This means that the only connection between Internet Identity and a wallet could be created only if the wallet imported the account from Internet Identity, which I don't think is possible in every case.

Yes, Internet Identity is not a wallet (yet) but only an identity provider that issues delegations (which allow non-interactive calls using the delegation identity). As such, it currently is not possible to do transaction approval on a call-by-call basis using Internet Identity.

> I believe that what we're defining here, through this wallet standard, is a secondary method of authentication in IC, an alternative to Internet Identity, where the source of the account can be any wallet that implements the standard. If that's the case, then the `permission` messages are indispensable as their main role is to obtain account related data from the wallet.

Yes, this is correct. I would not (yet) call the standard an alternative (as it has different goals and trade-offs) but rather a complement to it. The standards that we are defining here have transaction approval on a call-by-call basis as their core use-case. In the future, they can (and probably should) be extended to also incorporate additional use-cases such as issuing delegations. However, these extensions are definitely beyond the scope of RPF-7.
 
> It would be great, though, to hear from someone with a broader knowledge of the whole IC ecosystem to validate my thoughts 😅

I'm the original author of the consent message spec and RFP-7. Now that I'm back from my PTO I'm happy to answer questions, do reviews, etc. I believe @marydwyer (planned to?) set up a shared slack channel as well...
Feel free to reach out anytime. 🙂

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+    - If the user rejects the request, the wallet sends `error` back to the relying party with appropriate explanation.
+
+```
+{ // Request
+    type: "permission_request"
+    appMetadata: /* dApp */ {
+        name: String
+        icon: Optional<String>
+    }
+    networks: Array<{
```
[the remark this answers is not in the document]
jsamol: There is no public testnet, but you could set up a local one. I believe it could come useful to have wallets connect to a local replica, so I would refrain from removing this field entirely.

Thanks for bringing up the CAIP-2 specs, I wasn't aware of such a standard 😄 With that in light, I think we can replace `type` with `chainId` as both fields would effectively hold the same kind of information:
```
{
    chainId: string
    name?: string
    rpcUrl?: string
}
```
We still need `rpcUrl` to let the wallet know what host to use to make calls in case it's running locally, and `name` is there only for the UX.

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+        - basic information about the wallet (`appMetadata`),
+        - a list of networks on which the user has agreed the relying part can operate, should be a subset of the equivalent from `permission_request` (`networks`),
+        - a list of permission scopes the user has agreed the relying party can be granted, should be a subset of the equivalent from `permission_request` (`scopes`)
+        - the account details (princial or public key (*TBD what is required*) and optional subaccount) of the user selected account (`account`)
+    - If the user rejects the request, the wallet sends `error` back to the relying party with appropriate explanation.
+
+```
+{ // Request
+    type: "permission_request"
+    appMetadata: /* dApp */ {
```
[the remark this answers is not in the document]
jsamol: Yes, both sides create key pairs and exchange the public keys. The public key is the true source of identity which the user can use to verify that the request is coming from a dApp they trust. 

Normally, this public key is part of the message, but I omitted it deliberately for the first draft to focus solely on the purely IC-related content to make sure that we go in the right direction. Since I got my confirmation, I guess I can clutter it now with the technicalities 😅

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
frederikrothenberger: Good point about the local testnets. 👍 
I think it makes sense as suggested, thanks!

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
jsamol: Unfortunately, there's nothing stopping someone from creating a malicious copy of a popular dApp and try to completely pose as the original (UI wise, not only via internal channels of dApp and wallet communication). To prevent that, the wallet would have to know the dApp and its public key beforehand, meaning that it would have to use some kind of an allow list.

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+        rpcUrl: Optional<String>
+    }
+    canisterId: String
+    method: String
+    args: String /* encoded */ / Object /* decoded */
+}
+
+{ // Response
+    type: "canister_call_response"
+    response: String /* encoded */ / Object /* decoded */
```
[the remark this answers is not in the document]
jsamol: It would be returned in the response as if the call was successful. 

> I think it would be beneficial to specify that errors originating from the IC should be handed through to the relying party.

Yes, I agree. I think that thanks to there messages the wallet doesn't have to know the service it's calling, hence doesn't have to know how to decode the response. And if it doesn't know how to decode the response, it can't and shouldn't tell if it's successful or not 😄

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+        name: String
+        icon: Optional<String>
+    }
+    networks: Array<{
+        type: "mainnet" | "testnet"
+        name: Optional<String>
+        rpcUrl: Optional<String>
+    }>
+    scopes: Array<"canister_call">
+    account: {
```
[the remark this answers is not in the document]
jsamol: Something like
```
{ // Response
    type: "permission_response"
    appMetadata: /* wallet */ { ... }
    networks: Array<{ ... }>
    scopes: Array<...>
    principal: String
    ledger: Optional<{
        subaccount: Optional<String>
    }>
}
```
?

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+```
+
+## Canister Call
+
+Once the connection between the relying party and wallet is established, the relying party can request the wallet to execute canister calls:
+1. The relying party sends `canister_call_request`. The message incudes:
+    - the network details on which the call should be executed,
+    - the ID of the target canister,
```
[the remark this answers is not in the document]
jsamol: Oh, I believe that what I wrote [here](https://github.com/dfinity/wg-identity-authentication/pull/32#discussion_r1274920699) also applies here 😄 Let's discuss the need for the complete content map first in the other conversation, and I'll address it later, if I'm convinced it is required 😅

topics/dapp-wallet-msg.md
```diff
@@ -0,0 +1,110 @@
...
+    - the name of the call method to be executed,
+    - the arguments for the call.
+2. The wallet receives the request, validates if the relying party has permission to request the action and processes the message (in accordance with the ICRC-21 specification).
+    - If the user approves the request, the wallet sends the call to the IC and, upon receiving a successful response, sends `canister_call_response` back to the relying party. The message includes:
+        - the response from IC,
+        - optionally (*TBD*), the parameters (or their subset) of the request.
+    - If the user rejects the request or wallet fails to complete the requested action for any reason, the wallet sends `error` back to the relying party with appropriate explanation.
+
+```
+{ // Request
```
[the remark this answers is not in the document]
frederikrothenberger: > I've just noticed that the ICRC-21 draft says that the execution of the call is still done by the wallet and only the result of the call is handed back to the relying party, while the RFP-7 document specifies that the wallet is supposed to return only signatures for the call itself plus read_state that can come after.

Sorry for the discrepancy between RFP-7 and ICRC-21. I made some last minute changes to ICRC-21 after RFP-7 went out. We should go with the ICRC-21 variant (the wallet submits the call to the IC).
The changes to ICRC-21 were to address feedback from the internal product security team, that the wallet should be able to observe the interaction with IC in order to prevent malicious applications being able to lie to users about the outcome of submitting the signed call to the IC (i.e. one could imagine a malicious marketplace faking network loss in order to trick the user to sign a second payment transaction).

But in that model, for the relying party to verify that the wallet did not modify the request before submitting it to the IC (and to verify that the response is genuine), the relying party needs the full [certificate](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certificate) as returned by the `read_state` call and the full content map in order to calculate the [request id](https://internetcomputer.org/docs/current/references/ic-interface-spec/#request-id).

I have no strong opinion, whether the relying submits the full content map as part of the request or the wallet returns the matching content map back to the relying party together with the certificate, I think both could work fine.

> Regarding the types, I agree that they should be the same as the specification (arg and response should be bytes representing encoded values), but we have to keep in mind that the messages have to be serializable, sendable via different kind of transports, and readable on various platforms, therefore instead of using an actual byte array, I would use a hex string representation.

Good point! Is there an argument to be made against base64? This would make the strings much more compact.

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
frederikrothenberger: Yes, exactly. However, we need to be careful. Just because the wallet claims to have access to `principal` does not mean it actually does (hence why I proposed a challenge response scheme in the last working group session).

Alternatively, we could also reduce the set of information provided here (i.e. the wallet only gives feedback that the user approved the connection without giving information about the users identity). But this would make it very cumbersome for use-cases where the relying needs access to the principal in order to build the arguments for the request...

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
frederikrothenberger: One approach could also be to incorporate information provided by the browser. I.e. if the communication with the wallet is via `windowPostMessage` then the `origin` of the message is set by the browser and cannot be faked by a malicous dapp.
But this scheme only works over some channels unfortunately and is not a generic solution.

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
jsamol: > We should go with the ICRC-21 variant (the wallet submits the call to the IC).

👍 

> But in that model, for the relying party to verify that the wallet did not modify the request before submitting it to the IC (and to verify that the response is genuine), the relying party needs the full [certificate](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certificate) as returned by the read_state call and the full content map in order to calculate the [request id](https://internetcomputer.org/docs/current/references/ic-interface-spec/#request-id).

Just for me to understand, why should the relying party not trust the response from the wallet and need to verify that the request wasn't modified? 🤔 If the wallet was malicious and meant harm, the damage would be already done as the request would have been submitted before the relying party could verify anything. 

What's more, I always believed that *the relying party* is what should never be trusted, therefore the wallet (or rather the user through the wallet) should always have the last word in the requests, being even allowed to adjust them to the user's liking. A good example could be token transfers, where you can specify the fee you're willing to pay for the transaction. It wouldn't necessarily require a malicious act to send over a request with the fee set already and set to a value that the user finds inadequate and wants to change before submitting.

I can see that maybe from the UX perspective it could be useful to check if the response is what was expected to better communicate where the problem originated from, if there was any. But if the response was truly incorrect, it would get surfaced sooner or later while decoding, I think.

I mean, I'm not exactly advocating against such a model, but as a wallet developer, I find it quite complex and cumbersome (especially with the lack of support in the Dfinity libraries, but this is, I guess, something that could be fixed in the future releases) and I want to be sure I fully understand the reason behind such a requirement 😅 

> Is there an argument to be made against base64? This would make the strings much more compact.

Oh, you're right! Base64 should do a better job 😅

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
jsamol: > However, we need to be careful. Just because the wallet claims to have access to principal does not mean it actually does (hence why I proposed a challenge response scheme in the last working group session).

Ah, yes. Sorry, I forgot to include the challenge in my response, but I do remember about it 😄 

> Alternatively, we could also reduce the set of information provided here (i.e. the wallet only gives feedback that the user approved the connection without giving information about the users identity). But this would make it very cumbersome for use-cases where the relying needs access to the principal in order to build the arguments for the request...

From the UX point of view, users could want to know which identity they used to sign in to the relying parties, meaning that, most probably, the relying parties would always request to share the identity anyway. So unless there are some privacy-preserving concerns, I would keep the principal in the response. To keep it simple 😄

topics/dapp-wallet-msg.md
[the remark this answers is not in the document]
frederikrothenberger: It's about the relying party not trusting a specific wallet and the user. 

Imagine a gated application that only allows access if the user owns an asset with some specific canister. And furthermore, that canister does not offer a way to publicly query who owns assets.

In that case, the relying party would only need to trust that other canister and _not_ the wallet, if the response from the canister is handed through in such a way that the relying party can verify its authenticity.

In general, the wallet is under the users control and _should not_ be trusted anyway. I.e. if users can gain anything from using a wallet that does not forward genuine responses, they will absolutely use that.

And if the responses from the wallet are not verifiable, that just means that all interactions with the wallet need to be double checked by making out of band queries to the IC. Which also restricts the use cases to only those that actually can be checked publicly (since the relying party must not use the wallet for the check).

> What's more, I always believed that the relying party is what should never be trusted, therefore the wallet (or rather the user through the wallet) should always have the last word in the requests, being even allowed to adjust them to the user's liking.

I'm not advocating against allowing the user to tinker with the request within the wallet. This is useful, agreed. But the protocol should not support lying. 😉 

My point is that the relying party should receive all the information necessary to _verify_ the interaction with the IC. Having the wallet do A, but the tell the relying party it did B should lead to a protocol error (because the verification on the relying party side will fail).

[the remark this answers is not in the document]
jsamol: Ok, I think I understand now and I'm convinced 😄 I still keep forgetting that IC on many levels is different than most of the other networks and, for example, a lot of user data is not necessarily public.

> I have no strong opinion, whether the relying submits the full content map as part of the request or the wallet returns the matching content map back to the relying party together with the certificate, I think both could work fine.

I'd make that the wallet's responsibility to include the content map and certificate in the response, then. I think this should give more flexibility (if we can even speak of any 🤔) as the relying party could ask for a canister call result in the simplest way possible, let the wallet do its job, and only then worry about how the request was made and if the relying party is fine with it:

```
{ // Request
    type: "canister_call_request"
    network: {
        chainId: String
        name: Optional<String>
        rpcUrl: Optional<String>
    }
    canisterId: String
    method: String
    args: String /* encoded, base64 */
}

{ // Response
    type: "canister_call_response"
    network: {
         type: "mainnet" | "testnet"
         name: Optional<String>
         rpcUrl: Optional<String>
    }
    response: String /* encoded, base64 */
    contentMap: {
        request_type: String
        sender: String /* principal */
        nonce: Optional<String /* base64 */>
        ingress_expiry: Int
        canister_id: String /* base64 */
        method_name: String
        arg: String /* base64 */
    }
    certificate: {
        tree: // TODO: type
        signature: String /* base64 */
        delegation: // TODO: type
    }
}
```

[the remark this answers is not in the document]
frederikrothenberger: ah, another thing: The certificate already includes the response. So this response type duplicates that information. As such, I would only include the certificate (and remove the extra `response` field).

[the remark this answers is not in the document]
jsamol: > the network block on the response is still the old type.

Ah yes, thanks for pointing that out 😅 I corrected it.

Ok, I make sure to update the document before the next meeting.

---

> The certificate already includes the response. So this response type duplicates that information. As such, I would only include the certificate (and remove the extra response field).

Does it? Is it included in the tree?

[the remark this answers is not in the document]
frederikrothenberger: > Does it? Is it included in the tree?

Yes, see [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-request-status). The path to the response `blob` is `/request_status/<request_id>/reply`.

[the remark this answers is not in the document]
jsamol: I see! I wasn't aware 😅 I'll remove the `response` field right away

jsamol:
@frederikrothenberger I've updated the messages according to our discussion so far. I also added some technicalities not necessarily protocol specific, i.e. `version` and `senderId`. The `version` should allow us to seamlessly update the standard with new content in the future, and `senderId` is, well, the true identity of the source of the message that was discussed in one of the threads 😅 Additionally, I changed the form of the definitions from the original pseudocode to something more JSON-RPC suited.

One more thing regarding the `certificate` in the `canister_call` response. I've realized that the certificate returned by the `read_state` call is a CBOR encoded blob, not a raw object as I initially expected. Would it make sense, then, to change the type of the `certificate` field in the `canister_call` response to just `blob`?

frederikrothenberger:
Thanks a lot @jsamol! I'll try to take a look before tomorrows session.

> Would it make sense, then, to change the type of the certificate field in the canister_call response to just blob?

Yes, that is an option. Given the complexity already required to verify the certificate, having it as a cbor blob would not make it much harder to deal with.

topics/wallet-standard.md
```diff
@@ -0,0 +1,248 @@
...
+#### Types
+
+- `text`: A plain `string` value.
+- `blob`: A `string` value describing binary data encoded in base64.
+
+#### Request
+
+`version` (`text`): The version of the standard used.
+
+`senderId` (`text`): The ID of the sender. This should be used to identify the sender of the message.
```
frederikrothenberger, line 20: How is this different from `sender`? And what is the use-case for it?

I think a use-case example would help here. 😉  Also, if possible, another name might help too, since it is very similar to `sender` but describes something else?

topics/wallet-standard.md
```diff
@@ -0,0 +1,248 @@
...
+### Canister Call
+
+Once the connection between the relying party and wallet is established, the relying party can request the wallet to execute canister calls.
+
+#### Types
+
+- `text`: A plain `string` value.
+- `blob`: A `string` value describing binary data encoded in base64.
+- `nat`: A `string` value of an unsigned 64-bit integer.
+- `HashTree`: A recursive type defined as a union of:
```
frederikrothenberger, line 122: Looking at the definition of the type here, I think it might really be more practical to just treat certificate as a `blob`.

topics/wallet-standard.md
```diff
@@ -0,0 +1,248 @@
+# Wallet Standard
```
frederikrothenberger, line 1: I would call this a "Wallet _Message_ Standard" (or similar), since it only deals with this one aspect of wallets.

frederikrothenberger: I also took the liberty of [reserving an ICRC number (ICRC-25)](https://github.com/dfinity/ICRC/issues/29) for this specification. Please include it in the title.

topics/wallet-standard.md
```diff
@@ -0,0 +1,248 @@
...
+- `"canister_call"`
+
+`principal` (`blob`): The principal associated with the identity the user has selected.
+
+`ledger` (optional): Ledger data.
+- `subaccount` (`blob`, optional)
+
+`challenge` (`blob`): A challenge the wallet had to sign.
+
+`signature` (`blob`): The signature of the challenge generated by the wallet.
```
frederikrothenberger, line 61: I think we need some more details here: i.e. what signature schemes are allowed, how to validate the challenge, and how it relates to the principal.

And maybe some additional guidance on how a challenge should be chosen, etc.

topics/wallet-standard.md
```diff
@@ -0,0 +1,248 @@
...
+`contentMap`: The actual request content as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-call).
+- `request_type` (`text`)
+- `sender` (`blob`)
+- `nonce` (`blob`, optional)
+- `ingress_expiry` (`nat`)
+- `canister_id` (`blob`)
+- `method_name` (`text`)
+- `arg` (`blob`)
+
+`certificate`: The certificate returned by the `read_state` call as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certificate).
```
frederikrothenberger, line 167: We should probably also explain _why_ this is the certificate and how the relying party can verify that the response is genuine and matches the request. This can probably be just a few links into existing specifications.

frederikrothenberger (commented):
Let's see how the discussion in the working group goes, but to me this seems like it is almost done. Very nice! 👍 

The things left to do from my perspective:
* adjust the title
* complete the examples
* adjust some types
* specify the challenge mechanism more
* explain how the relying party can validate that the response matches the request

topics/wallet-standard.md
frederikrothenberger: Update 4: We agreed on naming it Wallet Interaction Standard.

topics/wallet-standard.md
jsamol: Since we agreed the`senderId` didn't belong here, it has been removed

jsamol:
@frederikrothenberger another draft update 😄
- I've updated the title
- I replaced the `principal` and `ledger` fields with one `identity` that holds, well, identity related details, including a DER-encoded `publicKey` + the crypto `algorithm` used to generate it (to make the `challenge` verification possible, as discussed) and an optional `ledger` (it felt like a good idea to group user-related data in the same object 🤔 )
- the `challenge` field in the `permission` request has a proper definition of what it is expected to be
- the `certificate` type in the `canister_call` response is `blob`
- I've updated the use cases with more detailed descriptions of how to process the messages (how to verify the challenge and `canister_call` response - I hope I got this part right 🤞)
- and finally, I've updated the examples and added a brief summary and terminology (since I use the same terminology as you do in ICRC-21, I took it from there)

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,265 @@
...
+- `ledger` (optional): Ledger data.
+    - `subaccount` (`blob`, optional)
+
+`challenge` (`blob`): The challenge the wallet had to sign.
+
+`signature` (`blob`): The signature produced by signing the challenge with the private key associated with the `identity`.
+
+#### Use-Case
+
+1. The relying party sends a `permission` request to the wallet.
```
frederikrothenberger, line 72: I would add a note that the `challenge` has to be generated from a secure random source.

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,265 @@
...
+3. If the response is successful, the relying party verifies whether the call performed by the wallet was genuine and retrieves the result:
+    - The relying party retrieves the `contentMap` from the response, verifies that its values match the expectations and uses it to [calculate a request id](https://internetcomputer.org/docs/current/references/ic-interface-spec/#request-id).
+    - The relying party retrieves the CBOR-encoded [`certificate`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certificate) from the response, decodes it and validates its authenticity with regard to [the root of trust](https://internetcomputer.org/docs/current/references/ic-interface-spec/#root-of-trust).
+        - If the validation process fails, the relying party rejects the response.
+    - The relying party uses the calulcated request id to retrieve [the `reply` blob](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-request-status) from the `certificate`'s state tree.
```
frederikrothenberger, line 190: Small typo: calulcated

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,265 @@
...
+- `arg` (`blob`)
+
+`certificate` (`blob`): The certificate returned by the `read_state` call as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certificate). The value is CBOR-encoded.
+
+#### Use-Case
+
+1. The relying party sends a `canister_call` request to the wallet.
+2. Upon receiving the request, the wallet validates whether the relying party has the permission to request the action and processes the message following the [ICRC-21]() specification.
+    - If the user approves the request:
+        - The wallet sends the call to the IC, retrieves its content map ([call](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-call), [query](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-query)) and [calculates a request id](https://internetcomputer.org/docs/current/references/ic-interface-spec/#request-id) based on it.
```
frederikrothenberger, line 181: Query calls work differently than update calls (i.e. they are synchronous, thus require no polling, but the result is also not certified).
Given that adding certification to query calls is application specific I would _not_ support signing `query` calls using the wallet.

Rather, it is possible to send messages to canisters as `update` even for canister methods marked as `query` in which case the call is executed in a replicated fashion and the result is certified as well (and verification works the same as for regular `update` calls).

So I would just add a note here that all calls (including queries) are sent as `update` calls.

frederikrothenberger (commented):
Nice work! Only a few minor comments left 👍

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,265 @@
...
+`networks`: A list of networks on which the user has agreed the relying party can operate. This should be a subset of the `networks` from the original request.
+- `chainId` (`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes that the user has agreed the relying party can be granted. This should be a subset of the `scopes` field from the original request. Possible values:
+- `"canister_call"`
+
+`identity`: The identity the user has selected.
+- `algorithm` (`text`): The cryptographic algorithm used to generate the key pair associated with the identity. The supported values correspond to signature schemes [supported by the IC](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures) and include:
```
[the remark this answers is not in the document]
frederikrothenberger: > we'll need more support from the dfinity libraries, though.

Yes, this is being worked on as we speak. There will be a rust library available that can be used to validate such signatures that is compatible with usage in IC canisters.

There could also be typescript bindings generated, however this is a bit further out.
The library should be available in a few weeks and I could probably bump the urgency if it becomes a blocker for the RFP.

> We could also go with the raw form instead as this would be a more versatile approach in the end.

More versatile how? It seems to me that we're reinventing the wheel if we go with a different approach. I would rather stick to established standards, for both compatibility and consistency.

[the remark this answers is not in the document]
jsamol: > Yes, this is being worked on as we speak. There will be a rust library available that can be used to validate such signatures that is compatible with usage in IC canisters.

❤️ 

> There could also be typescript bindings generated, however this is a bit further out.
The library should be available in a few weeks and I could probably bump the urgency if it becomes a blocker for the RFP.

It's not a blocker, but it would be nice to have tools that help seamlessly construct and process the messages around the time the standard gets published 😅 

I updated the document 👍

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,254 @@
...
+`signature` (`blob`): The signature produced by signing the challenge with the private key associated with the `identity`.
+
+#### Use-Case
+
+1. The relying party sends a `permission` request to the wallet.
+2. Upon receiving the message, the wallet presents the details of the to-be-established connection to the user and asks the user to select the identity that will be paired in response.
+    - If the user approves the request, the wallet sends a succesful response back to the relying party.
+    - If the user rejects the request, the wallet sends a response with an error back to the relying party.
+3. After receiving a succesful response, the relying party verifies that the wallet has access to the private key associated with the provided identity:
+    - The relying party retrieves the `publicKey` from `identity`, determines the `signature` scheme and verifies whether it was generated by signing the `challenge` from the request with the private key associated with the `publicKey`.
```
frederikrothenberger, line 71: I just realized, we also need to specify a domain separator (see [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures)), to make sure that the signature cannot be misused in other contexts.

My suggestion would be something like `\x0Aic-wallet-challenge` to be consistent with other domain separators specified in the interface spec (examples in this [section](https://internetcomputer.org/docs/current/references/ic-interface-spec/#authentication)).

jsamol: Good point 👍 I've updated that part

sea-snake:
> including a DER-encoded publicKey + the crypto algorithm used to generate it (to make the challenge verification possible, as discussed)

The whole idea behind DER encoding is that it already includes the algorithm of the public key.

sea-snake:
Here's how I verify challenges signed with either `ECDSAKeyIdentity`, `Ed25519KeyIdentity` or `Secp256k1KeyIdentity` with only the DER-encoded public key:
```
import { p256 } from '@noble/curves/p256';
import { secp256k1 } from '@noble/curves/secp256k1';
import * as ed from '@noble/ed25519';
import { sha512 } from '@noble/hashes/sha512';
import { sha256 } from '@noble/hashes/sha256';
import { ED25519_OID, unwrapDER } from '@dfinity/identity';

ed.etc.sha512Sync = (...m) => sha512(ed.etc.concatBytes(...m));

const ECDSA_P256_OID = Uint8Array.from([
  ...[0x30, 0x13], // SEQUENCE
  ...[0x06, 0x07], // OID with 7 bytes
  ...[0x2a, 0x86, 0x48, 0xce, 0x3d, 0x02, 0x01], // OID ECDSA
  ...[0x06, 0x08], // OID with 8 bytes
  ...[0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07], // OID P-256
]);

const SECP256K1_OID = Uint8Array.from([
  ...[0x30, 0x10], // SEQUENCE
  ...[0x06, 0x07], // OID with 7 bytes
  ...[0x2a, 0x86, 0x48, 0xce, 0x3d, 0x02, 0x01], // OID ECDSA
  ...[0x06, 0x05], // OID with 5 bytes
  ...[0x2b, 0x81, 0x04, 0x00, 0x0a], // OID secp256k1
]);

export const verifyChallenge = async (
  publicKey: Uint8Array,
  signature: Uint8Array,
  data: Uint8Array,
): Promise<boolean> => {
  try {
    if (publicKey[0] !== 0x30 || publicKey[2] !== 0x30) {
      return false;
    }
    const oidSequenceLength = publicKey[3];
    const oid = publicKey.slice(2, oidSequenceLength + 4);
    if (oid.length !== oidSequenceLength + 2) {
      return false;
    }
    switch (Array.from(oid).join()) {
      case Array.from(ECDSA_P256_OID).join():
        return p256.verify(
          Buffer.from(signature).toString('hex'),
          Buffer.from(sha256(data)).toString('hex'),
          Buffer.from(unwrapDER(publicKey, ECDSA_P256_OID)).toString('hex'),
        );
      case Array.from(ED25519_OID).join():
        return ed.verify(
          Buffer.from(signature).toString('hex'),
          Buffer.from(data).toString('hex'),
          Buffer.from(unwrapDER(publicKey, ED25519_OID)).toString('hex'),
        );
      case Array.from(SECP256K1_OID).join():
        return secp256k1.verify(
          Buffer.from(signature).toString('hex'),
          Buffer.from(sha256(data)).toString('hex'),
          Buffer.from(unwrapDER(publicKey, SECP256K1_OID)).toString('hex'),
        );
      default:
        return false;
    }
  } catch (_) {
    throw Error;
  }
};
```

The DER-encoding actually standardizes the way the key and algorithm are communicated.

frederikrothenberger:
> The whole idea behind DER encoding is that it already includes the algorithm of the public key.

@sea-snake: Thanks! The latest version does no longer include the crypto algorithm separately exactly because of that. 👍

jsamol:
> Here's how I verify challenges signed with either ECDSAKeyIdentity, Ed25519KeyIdentity or Secp256k1KeyIdentity with only the DER-encoded public key: [...]
@sea-snake thanks for the snippet, it's very helpful! For the best development experience, however, I would love to see this functionality to be part of the `@dfinity` libraries, especially that there's already the `PublicKey` abstraction which could be extended with a factory function, which constructs a `PublicKey` implementation from a DER-encoded value, and with a `verify` method, which verifies the signature depending on the implementation.

sea-snake:
@jsamol When I find the time, I can make an MR to make that happen. Not sure what the impact is on the package size, might need to be it's own lib.

sea-snake:
> The wallet continues to call `read_state` for the calculated request id until [the status of the call](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-request-status) indicates that the call has been processed (succesfully or not).

I don't see why the wallet should be making any requests and even poll for the result which can take some time. Can't the wallet just return the signed canister call and signed read state request so that the dapp can execute those however it intends?
This is basically how I've currently implemented things currently.

frederikrothenberger:
> > The wallet continues to call `read_state` for the calculated request id until [the status of the call](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-request-status) indicates that the call has been processed (succesfully or not).
> 
> I don't see why the wallet should be making any requests and even poll for the result which can take some time. Can't the wallet just return the signed canister call and signed read state request so that the dapp can execute those however it intends? This is basically how I've currently implemented things, the wallet is a popup that closes after returning a response over postMessage.

This was the initial design. However, a DFINITY internal security review found that it would be safer to do it this way. The reason was the following attack scenario:
* The canister call is not idempotent
* The result of the canister call is not publicly observable

In that case a malicious relying party can request a transaction from the wallet. Given it is the relying party that submits the call, the wallet will _not_ observe the result of the call. A malicious relying party could successfully submit the call but then show some made up error to the user tricking them into repeating the transaction. Having the wallet submit the call (and thus is able to observe the result) makes this type of attack more difficult to pull off (as the wallet would already show a success message to the user).

frederikrothenberger:
> > A challenge used for the wallet to sign in order to prove its access to the identity. The challenge should be an array of 32 cryptographically random bytes generated from a secure random source by the sender of the request.
> 
> If the wallet were to delegate, it could sign a delegation instead of a challenge, see example repo above.
> 
> Using delegations and having a rpc communication channel over url/postMessage/callback besides browser extensions that inject an rpc service, enables wallets that are hosted on the IC without a browser extension to function like NFID.

I think there is a use-case to exchange delegation over such a channel using JSON rpc. However, I would make this an extension of this standard by extending it by introducing an additional scope (e.g. "delegation"), alongside with a delegation request and response.

The reason why I would split it into a separate standard is that the currently proposed feature set is save over any channel, whereas exchanging delegations usually requires an _authenticated_ channel.

sea-snake:
> * The canister call is not idempotent

I remember asking about this a few times before, as far as I've been told, you can't submit a canister call with the same request id more than once. So I don't see a security issue here.

> The result of the canister call is not publicly observable
That's a good point, sadly that means the UX for the end user isn't that great (waiting for a popup to close since the call is still ongoing).

sea-snake:
> The reason why I would split it into a separate standard is that the currently proposed feature set is save over any channel, whereas exchanging delegations usually requires an _authenticated_ channel.

What do you mean with authenticated? As in data could be intercepted if it's not? A delegation itself isn't something that needs to be secured as long the as private key the delegation is created for isn't communicated over any channel (which it shouldn't in the first place).

frederikrothenberger:
>That's a good point, sadly that means the UX for the end user isn't that great (waiting for a popup to close since the call is still ongoing).

I don't think it is that bad. If the user has to look at a spinner for 2 seconds, that's still very manageable.

sea-snake:
> All security models around delegations that I have seen so far require this property, because either the delegation principal or the target list is dependent on the sender of the delegation request.

Yeah that's why I require a public key in the initial connection that looks at the origin. But subsequent calls can be made by signing a challenge over the request id, the origin doesn't need to be checked anymore. So that the dapp can for example connect at http://account.dapp.com while making calls from http://market.dapp.com by simply sharing the private key with it's own domains.

sea-snake:
I'll implement an Agent following this spec and link it here :D

frederikrothenberger:
> > Upon receiving the request, the wallet validates whether the relying party has the permission to request the action and processes the message following the ICRC-21 specification.
> 
> This would limit the dapp from only being able to make calls to canisters that have this implemented. Is it fair to assume this is optional, the wallet can also decide to decode the canister call arguments and show these to the user?

Yes, of course. However, it is recommended that some sort of warning is shown to the user to make them aware that this target does not fully support transaction approval.

frederikrothenberger:
> > All security models around delegations that I have seen so far require this property, because either the delegation principal or the target list is dependent on the sender of the delegation request.
> 
> Yeah that's why I require a public key in the initial connection that looks at the origin. But subsequent calls can be made by signing a challenge over the request id, the origin doesn't need to be checked anymore. So that the dapp can for example connect at http://account.dapp.com while making calls from http://market.dapp.com by simply sharing the private key with it's own domains.

Ah, I see. I think this model of sharing the "sender identity" with another domain could be an extension too, right?
I'd like to keep the standards small & focused in order to not burden devs too much that don't need the full functionality.
Or do you see a fundamental issue with ICRC-25 that would break with what you are trying to do?

sea-snake:
I've ran into an issue while I was implementing an `Agent` for the spec.

> The relying party retrieves the `publicKey` from `identity`, determines the `signature` scheme and verifies whether it was generated by signing the concatenation of the domain separator `\x0Aic-wallet-challenge` (UTF-8 encoded) and the `challenge` from the request with the private key associated with the `publicKey`.

This isn't possible if the wallet doesn't have access to the private key of the identity, which actually if the case with Internet Identity and wallets that rely on delegations to secure their private key behind biometrics/password/non-extractable-browser-key. Both is the case for the [wallet](https://etk52-fqaaa-aaaak-ae4ca-cai.icp0.io/) I'm currently implementing, and there's no way around it 😅

Why does the relying party need to verify that the wallet has the private key for the public key? Only reason I can think of would be to verify it for e.g. user registration, but even then I would probably rely on a canister call instead.

One workaround I could think of would be to also return a delegation besides the signed challenge. If a delegation is returned, the challenge can be verified with the public key at the end of the delegation chain and then the delegation chain itself can be validated to be valid & start with the identity public key.

Adding an optional delegation chain to the identity object in the response would make above doable, though it makes things more complex, it also makes it possible to work with identities beyond the ones that directly come from a seed phrase.

frederikrothenberger:
> Why does the relying party need to verify that the wallet has the private key for the public key? Only reason I can think of would be to verify it for e.g. user registration, but even then I would probably rely on a canister call instead.

The main reason is to prevent wallets being able to claim control of a key they don't actually have access to. I expect quite a few applications building a poor mans sign-in flow where the signed challenge serves as some sort of session token. Having the challenge response flow at least makes sure that the wallet cannot lie about the keys it controls.

Having the option to provide a delegation chain seems sensible to me. While it does add complexity, it is also a largely solved problem (there are libraries for that) and consistent with the ecosystem.

What do you think?

PS: It would be really nice if you could join the working group session on September 3 to discuss your findings and feedback. 🙂

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,255 @@
...
+- `name` (`text`): The user-friendly name of the relying party. 
+- `icon` (`text`, optional): An optional URL pointing to an icon resource representing the relying party.
+
+`networks`: A list of networks on which the relying party plans to operate.
+- `chainId`(`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes the relying party requires. Possible values: 
+- `"canister_call"`
```
frederikrothenberger, line 37: To have this standard be "extension proof" I would explicitly state that a wallet should ignore scope values it does not recognize.

This keeps the door open to later introduce additional values in other standards (as an extension), while wallets that only implement this standard have a well defined way of handling the extension scopes.

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,255 @@
...
+
+`networks`: A list of networks on which the user has agreed the relying party can operate. This should be a subset of the `networks` from the original request.
+- `chainId` (`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes that the user has agreed the relying party can be granted. This should be a subset of the `scopes` field from the original request. Possible values:
+- `"canister_call"`
+
+`identity`: The identity the user has selected.
```
frederikrothenberger, line 57: I think there is a use-case for sharing multiple keys at once. Imagine having a dapp that specializes in asset management. It would be nice if you could easily plug in your wallet that controls multiple keys into that dapp to control all your assets (spread across many keys) with a single connection request.

So my suggestion is to change `identity` to `identities` (array) containing objects of the same shape as the current `identity`.

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,255 @@
...
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity. The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
+- `ledger` (optional): Ledger data.
+    - `subaccount` (`blob`, optional)
```
frederikrothenberger, line 60: I think `subaccount` should be changed to an array of `blob` as well, since you can easily have many.

topics/wallet-interaction-standard.md
sea-snake: Yeah this also makes sense if the wallet derives multiple private keys instead of subaccounts like Plug does. Do we want to have optional names for identities?

topics/wallet-interaction-standard.md
sea-snake: Yeah this makes sense too. Do we also want to have optional names for these?

topics/wallet-interaction-standard.md
frederikrothenberger: Hm, I'm currently leaning towards having key names be a wallet internal concept. The wallet should of course show the key name on the transaction approval screen.

But I have no strong opinion on that. As an optional field, I think it's good to have. Then it is still up to the wallet whether or not to provide it (or make it a user choice). 👍

sea-snake:
> The main reason is to prevent wallets being able to claim control of a key they don't actually have access to. I expect quite a few applications building a poor mans sign-in flow where the signed challenge serves as some sort of session token. Having the challenge response flow at least makes sure that the wallet cannot lie about the keys it controls.

That's... terrifying but I totally see that happen indeed, especially with web2 sites that use a web3 wallet.

> Having the option to provide a delegation chain seems sensible to me. While it does add complexity, it is also a largely solved problem (there are libraries for that) and consistent with the ecosystem.
> 
> What do you think?

I'll add the delegation_chain as an optional response value within the identity object in my implementation.

> PS: It would be really nice if you could join the working group session on September 3 to discuss your findings and feedback. 🙂

I'll be there.

sea-snake:
@frederikrothenberger 
In the @dfinity/agent lib the delegation chain is converted to JSON in a different way than this spec, public key blobs are encoded as hex instead of base64 and targets (principals) are also encoded as hex instead base32 dash separated string.

https://github.com/dfinity/agent-js/blob/main/packages/identity/src/identity/delegation.ts#L48

So should I ignore that and follow the encodings in this spec?

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,255 @@
...
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes the relying party requires. Possible values: 
+- `"canister_call"`
+
+`challenge` (`blob`): A challenge used for the wallet to sign in order to prove its access to the identity. The challenge should be an array of 32 cryptographically random bytes generated from a secure random source by the sender of the request.
+
+#### Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
```
sea-snake, line 43: Version is defined as text here but in the example request it's a number? Looking at the JSON-RPC 2.0 spec, numbers are valid values besides text but things bigint will need to be encoded as string.

frederikrothenberger:
> @frederikrothenberger In the @dfinity/agent lib the delegation chain is converted to JSON in a different way than this spec, public key blobs are encoded as hex instead of base64 and targets (principals) are also encoded as hex instead base32 dash separated string.
> 
> https://github.com/dfinity/agent-js/blob/main/packages/identity/src/identity/delegation.ts#L48
> 
> So should I ignore that and follow the encodings in this spec? Probably makes sense anyway since the public key within the delegation chain is redundant (already defined within identity).

Hm, that's true. Yeah, it is probably better to be consistent with existing code. Let's pick this up as a discussion point in the next session...

topics/wallet-interaction-standard.md
jsamol: I just went with the most versatile type here to open it to values other than just numbers, for example two- or three-part version numbers, or something more descriptive like, for example, `'1.icrcXX'` for extensions of this standard. However, I'm not sure, if there's a real use case for such versions 🤔 if we were to stick to just a single number versions, then I agree the type could be narrowed down to just a number.

sea-snake:
Took a bit longer than expected, but here's the `WalletAgent` implementation that follows this spec: https://github.com/slide-computer/identity/blob/master/src/index.ts

It's definitely a bit of a chaotic code trying to wrap things within the constraints of the `Agent` interface, while trying to support the `pollForResponse` implementation, upgrading query requests to calls and adding support for delegation chains.

This agent should be interchangeable with `HttpAgent` when you create an `Actor`.

As for the data transport, this is a configurable option when you create the agent, the following transports have been implemented: `LinkTransport`, `PostMessageTransport`, `PromiseTransport` and `MixedTransport` and more implementation could be added

Maybe a standard that lets you lookup the recommended transport and its options for a specific wallet and device environment, could simplify things here further into something like a `LookupTransport` implementation that only requires the wallet url to get things working.

frederikrothenberger:
@sea-snake: That's fantastic! 😄 Do you have a deployment of your code somewhere, so that I can try it out in a live example?

> Maybe a standard that lets you lookup the recommended transport and its options for a specific wallet and device environment, could simplify things here further into something like a LookupTransport implementation that only requires the wallet url to get things working.

This is absolutely something we should work on once the standard has been established. As I see it, the next steps are as follows:
1. Standardize ICRC-21 and ICRC-25
2. Outline possible extensions / further work.
   Current candidates are:
    * Standardize transports & transport selection
    * Support exchanging delegations by introducing the `delegation` scope (and another request / response pair)
    * Extend ICRC-25 to support batch approvals
3. Work on extensions

sea-snake:
>  Do you have a deployment of your code somewhere, so that I can try it out in a live example?

Working on that :D

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,257 @@
...
+`networks`: A list of networks on which the user has agreed the relying party can operate. This should be a subset of the `networks` from the original request.
+- `chainId` (`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes that the wallet supports and the user has agreed the relying party can be granted. This should be a subset of the `scopes` field from the original request. Possible values:
+- `"canister_call"`
+
+`identities`: A list of identities the user has selected.
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity. The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
```
frederikrothenberger, line 58: Could we add an optional `name` for the entries of `identities`?

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,257 @@
...
+- `ledger` (optional): Ledger data.
+    - `subaccounts` (`blob`, optional): A list of subaccounts the user has selected.
```
frederikrothenberger, line 60: Could we add an optional `name` for the entries of `subaccounts`?

sea-snake: @frederikrothenberger Where would we put the blob when subaccounts becomes and object? I currently implemented this structure:
```ts
subAccount: {
    bytes: blob;
    name?: string;
}
```

frederikrothenberger: Yes, but it's an array of objects, so:
```
subAccounts: {
    bytes: blob;
    name?: string;
}[]
```

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,257 @@
...
+- `chainId` (`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`scopes`: A list of permission scopes that the wallet supports and the user has agreed the relying party can be granted. This should be a subset of the `scopes` field from the original request. Possible values:
+- `"canister_call"`
+
+`identities`: A list of identities the user has selected.
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity. The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
+- `ledger` (optional): Ledger data.
```
frederikrothenberger, line 59: We should make it clear that this refers to **ICRC-1** ledgers.

topics/wallet-interaction-standard.md
frederikrothenberger: I think that is consistent with how accounts are defined in [ICRC-1](https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1/README.md):
`type Account = record { owner : principal; subaccount : opt Subaccount; };`
I.e. you can leave it off completely.

topics/wallet-interaction-standard.md
frederikrothenberger, line 58: We also need an optional list of delegations here.

neeboo:
> > Do you have a deployment of your code somewhere, so that I can try it out in a live example?
> 
> Working on that :D

Please provide benchmark btw :D

frederikrothenberger:
> > > Do you have a deployment of your code somewhere, so that I can try it out in a live example?
> > 
> > 
> > Working on that :D
> 
> Please provide benchmark btw :D

@neeboo: could you please clarify what metrics you are interested in?

neeboo:
> > > > Do you have a deployment of your code somewhere, so that I can try it out in a live example?
> > > 
> > > 
> > > Working on that :D
> > 
> > 
> > Please provide benchmark btw :D
> 
> @neeboo: could you please clarify what metrics you are interested in?

mainly latency.

in the code, `agent.permission` is an await, and `ledger.transfer` is another.

And also bulk transfer is something dapp devs really want, would it be considered as well?

frederikrothenberger:
> And also bulk transfer is something dapp devs really want, would it be considered as well?

Yes, batching / bulk transfer is probably a natural extension of this standard. See my comment [here](https://github.com/dfinity/wg-identity-authentication/pull/32#issuecomment-1698595687).

frederikrothenberger:
> mainly latency.
>
> in the code, agent.permission is an await, and ledger.transfer is another.

Note that in both of these interactions a user approval is required. It is generally expected that the user interaction with the wallet will make up the bulk of any latency measurement.

neeboo:
> > mainly latency.
> > in the code, agent.permission is an await, and ledger.transfer is another.
> 
> Note that in both of these interactions a user approval is required. It is generally expected that the user interaction with the wallet will make up the bulk of any latency measurement.

yep, i think it really depends on wallet to handle interface and etc.

sea-snake:
> > > > > Do you have a deployment of your code somewhere, so that I can try it out in a live example?
> > > > 
> > > > 
> > > > Working on that :D
> > > 
> > > 
> > > Please provide benchmark btw :D
> > 
> > 
> > @neeboo: could you please clarify what metrics you are interested in?
> 
> mainly latency.
> 
> in the code, `agent.permission` is an await, and `ledger.transfer` is another.
> 
> And also bulk transfer is something dapp devs really want, would it be considered as well?

Permission is a one time operation, the response data can be stored in the browser so the dapp knows it already has permission and reuses this data until it receives a no permissions error when the user manually revoked the permission in the wallet.

Bulk is primarily useful for multiple operations that the user needs to approve towards ledgers and other dapps.

More common are bulk calls from a dapp frontent to the dapp backend canister(s), with delegations you can avoid going through the wallet altogether and directly make those calls from the dapp frontend.

neeboo:
> > > > > > Do you have a deployment of your code somewhere, so that I can try it out in a live example?
> > > > > 
> > > > > 
> > > > > Working on that :D
> > > > 
> > > > 
> > > > Please provide benchmark btw :D
> > > 
> > > 
> > > @neeboo: could you please clarify what metrics you are interested in?
> > 
> > 
> > mainly latency.
> > in the code, `agent.permission` is an await, and `ledger.transfer` is another.
> > And also bulk transfer is something dapp devs really want, would it be considered as well?
> 
> Permission is a one time operation, the response data can be stored in the browser so the dapp knows it already has permission and reuses this data until it receives a no permissions error when the user manually revoked the permission in the wallet.
> 
> Bulk is primarily useful for multiple operations that the user needs to approve towards ledgers and other dapps.
> 
> More common are bulk calls from a dapp frontent to the dapp backend canister(s), with delegations you can avoid going through the wallet altogether and directly make those calls from the dapp frontend.

Thanks for the reply. Personally, I really want to see how UI handles those messages and make permission smoothly.
Can't wait to see your example.

sea-snake:
Based on [forum discussion](https://forum.dfinity.org/t/call-for-feedback-ic-wallet-standards/22489) there's basically one main point of discussion, the choice of JSON RPC for the spec. 

As I mentioned in a forum post, maybe Candid IDL makes more sense to be platform agnostic and then a spec could be defined for Candid -> JSON RPC for data transport channels where it makes more sense then Candid.

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,257 @@
...
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`network`: Network details on which the call was executed.
+- `chainId` (`text`): The chain id of the network as described in the [CAIP-2]() standard.
+- `name` (`text`, optional): An optional user-friendly name of the network.
+- `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
+
+`contentMap`: The actual request content as specified [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-call).
+- `request_type` (`text`)
+- `sender` (`blob`)
```
sea-snake, line 158: I've seen a mix of sender and canister id defined as both blob and text, we should probably be consistent here. In the example principals have are sent as their base32 string not as their base64 encoded bytes, so that would mean text instead of blob.

sea-snake:
Here's a working demo: https://osu4t-5aaaa-aaaak-ae5za-cai.icp0.io

Make sure to create a wallet first at: https://etk52-fqaaa-aaaak-ae4ca-cai.icp0.io

The wallet and this demo are still WIP:

- The wallet and specifically demo are not fully mobile optimized yet
- Error handling is not fully implemented, it's basically missing in the demo
- Make sure to be signed in to the wallet, if you're not, things will break in this demo
- Secure signing and locking settings do not work yet in this demo flow
- The wallet does not remember the permission, you'll have to reconnect every time (not implemented yet)
- Delegations work but is currently hardcoded in the wallet to a cowsay canister for the purpose of this demo
- Internet Identity works in the wallet and this demo, but if it expires, the wallet won't let you know yet 😅

Functionality in this demo:
- Connect to wallet by permission request
- Transfer ICP by canister_call request
- Query cowsay canister by delegation (no interaction or communication with wallet required)

topics/wallet-interaction-standard.md
```diff
@@ -0,0 +1,257 @@
...
+
+`scopes`: A list of permission scopes that the wallet supports and the user has agreed the relying party can be granted. This should be a subset of the `scopes` field from the original request. Possible values:
+- `"canister_call"`
+
+`identities`: A list of identities the user has selected.
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity. The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
+- `ledger` (optional): Ledger data.
+    - `subaccounts` (`blob`, optional): A list of subaccounts the user has selected.
+
+`signature` (`blob`): The signature produced by signing the concatenation of the domain separator `\x0Aic-wallet-challenge` (UTF-8 encoded) and the challenge with the private key associated with the `identity`.
```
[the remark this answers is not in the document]
sea-snake: > I made a mistake here, the domain separator should be `\x13ic-wallet-challenge`, as the first byte must indicate the length of the remainder of the domain separator.

I always wondered what the purpose of those first byte chars were in domain separators, so it's basically like the sequence length byte in a der key.

jsamol:
PoC (up to date with the standard definition at `a03d16d6400abc18666995009785e56cd1235c52`):
- DApp: https://weutz-riaaa-aaaao-a2hna-cai.icp0.io/
- Wallet: https://wdvvn-4qaaa-aaaao-a2hnq-cai.icp0.io/
- ICRC-21 canister: https://dashboard.internetcomputer.org/canister/xhy27-fqaaa-aaaao-a2hlq-cai

frederikrothenberger (approved):

frederikrothenberger:
The PR should be merged so that the draft can be worked on in smaller, focused PRs.

PR #36, opened 2023-08-31
ICRC-21: Add not_supported error to enable gradual roll-out

frederikrothenberger:
This PR introduces an additional error to give developers a way of expressing that support is not there (yet).


przydatek (approved):
LGTM, thanks!

PR #39, opened 2023-09-04
Prefix ICRC-21 methods and types

frederikrothenberger:
In order to separate the ICRC-21 mandated types and methods from other methods and types a canister might use, they should be prefixed with the standard name "icrc21_".

This PR introduces these prefixes. It also fixes a small inconsistency in the example.


przydatek (approved):
LGTM.

PR #40, opened 2023-09-05
Introduce ICRC-21 supported standards query

frederikrothenberger:
This adds a query method to the ICRC-21 canister interface to list the supported standards related to consent messages.


przydatek (approved):
LGTM.

PR #45, opened 2023-09-06
ICRC-25: Remove app metadata

frederikrothenberger:
This PR removes app metadata from the permission request.

Reason: the standard currently gives no way to verify the supplied information. Any way of verifying the app metadata is dependent on the transport used. For this reason, the exchange of `appMetadata` should be handled by the transport standard.


jsamol:
I believe you also should remove `walletMetadata` from the [response](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/wallet-interaction-standard.md#response). I forgot we had opted for removing it and renamed it according to the feedback 😅

frederikrothenberger:
@jsamol: I have now removed the wallet metadata too.

jsamol (approved):

PR #55, opened 2023-09-27
Rename ICRC-25 to use signer instead of wallet

frederikrothenberger:
This renaming makes the terminology more specific.
"Wallet" is an overloaded term and implies asset management
functionality (e.g. token listings & transfers) for some people.

As no such functionality is required by ICRC-25 we call it a signer
now (because the component is required to sign things).

In addition, the file is renamed to contain the ICRC number.


przydatek (approved):
LGTM, thanks!

Issue #52, opened 2023-09-25
ICRC-25: Security Review, Resolution of Findings 


frederikrothenberger:
See the attached security review: [2023-09 ICRC-25 Wallet Interaction Design Review.pdf](https://github.com/dfinity/wg-identity-authentication/files/12712447/2023-09.ICRC-25.Wallet.Interaction.Design.Review.pdf)

This issue will be used to track the resolution of the findings:
* [x] F01: Wallet should warn about unrecognized RPs
* [x] F02: Security Requirements Are Underspecified
* [x] F03: Permission scopes are too broad
* [x] F04: Standard neglects permission deletion
* [x] F05: RP is not required to verify failed calls to the target canister
* [x] F06: Authentication of RP relies on DNS and centralized authorities

frederikrothenberger:
F01 addressed in #58.

frederikrothenberger:
F03 addressed in #60.

frederikrothenberger:
F05 addressed in #64.

frederikrothenberger:
F06 addressed in #62, by making it a property of the transport channel.

frederikrothenberger:
All security findings have been addressed.

PR #60, opened 2023-09-28
ICRC-25: Make permission scope extensible, addresses sec finding F03

frederikrothenberger:
This PR addresses the [security review](https://github.com/dfinity/wg-identity-authentication/issues/52) finding F03 by making the permission scope extensible. This allows to define additional properties of the permission scope object in the future, such as restricting the `canister_call` scope to just certain canisters, methods, or even arguments.

In addition, it also makes a formulation stricter that would have allowed signers to hand out more permissions than initially requested.

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -38,7 +38,7 @@ The purpose of the `permission` messages is to establish a connection between a 
 - `name` (`text`, optional): An optional user-friendly name of the network.
 - `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
 
-`scopes`: A list of permission scope objects the relying party requires. If the signer does not support any of the requested scopes, it should ignore it. Permission scope properties:
+`scopes`: A list of permission scope objects the relying party requires. If the signer does not support a requested scope, it should ignore that particular scope and proceed as if the `scopes` list did not include that object. Permission scope properties:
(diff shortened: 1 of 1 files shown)


topics/icrc_25_signer_interaction_standard.md
```diff
@@ -38,8 +38,8 @@ The purpose of the `permission` messages is to establish a connection between a
 - `name` (`text`, optional): An optional user-friendly name of the network.
 - `rpcUrl` (`text`, optional): An optional custom RPC URL associated with the network.
 
-`scopes`: A list of permission scopes the relying party requires. If the signer does not support any of the requested scopes, it should ignore it. Possible values: 
-- `"canister_call"`
+`scopes`: A list of permission scope objects the relying party requires. If the signer does not support any of the requested scopes, it should ignore it. Permission scope properties:
```
przydatek, line 41: This is not about the change in this PR, but in the sentence "If the signer does not support any of the requested scopes, it should ignore it." the final "it" that should be ignored is not clear: should the signer ignore the request (i.e. not reply at all)?, or should the signer reply with an error?

przydatek (approved):
LGTM, modulo a clarifying question.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Thanks, I reformulated the sentence.

PR #61, opened 2023-09-29
ICRC-21: Adapt the authentication requirements for icrc21_consent_message

frederikrothenberger:
This PR changes the authentication requirements for the icrc21_consent_message method call:
* authentication is no longer mandatory: this improves the flow significantly for cold signers
* authentication may still give benefits, if supported by the target canister. This could potentially lead to better consent messages for hot signer that can make this call in an authenticated way.

Additionally, add a note about the upcoming change for `icrc21_consent_message` which changes it from an `update` call to a relicated query as soon as that feature is generally available.


przydatek (approved):
LGTM, thanks!

PR #63, opened 2023-10-02
ICRC-25 reference implementation

jsamol:
This PR adds a reference implementation for the ICRC-25 standard. It includes:
- an example relying party
- an example signer
- a simple target canister implementing the ICRC-21 standard

The implementation in its current state is outdated and is based on the draft version from #32.


reference-implementations/icrc25/canister/dfx.json
```diff
@@ -0,0 +1,46 @@
+{
+  "canisters": {
+    "ledger": {
+      "type": "custom",
+      "candid": "https://raw.githubusercontent.com/dfinity/ic/a4191e757dbc581cbc8cef9160aea1e852218a1a/rs/rosetta-api/icrc1/ledger/ledger.did",
```
frederikrothenberger, line 5: I think this approach here is sufficient to reference a ledger canister. I would drop the `dfx.ic.json` together with the contents of the `icrc1_ledger` folder.

frederikrothenberger (commented):
I think the following things can still be removed:
* icrc_21 front-end including the webpack config, package.json, dfx.json entries
* Everything related to mainnet: e.g. `canister_ids.json` files
* Everything not strictly needed in the front-ends. These are dummy front-ends and simplicity is the most important thing (right after actually implementing the standard). So I would suggest to drop the `robots.txt`, `manifest.json`, logos in various sizes...
* While I think the reference implementation should have tests, I'm not sure the current dummy tests and their setup contribute much. The standards are about interactions between web-apps, so we might need to have selenium tests anyway that operate across the relying party and the signer. For now I would simply remove the test stubs.
* Rework the readmes, so that they do not contain generated default content. Also add a top-level readme that explains the project structure.

frederikrothenberger:
Thanks a lot for all the work! It still needs some polishing, but I think it is a good start. 🙂

jsamol:
@frederikrothenberger I updated the PR according to the feedback, I hope I haven't missed anything.

reference-implementations/icrc25/README.md
```diff
@@ -0,0 +1,38 @@
+# ICRC-25 Reference Implementation
```
frederikrothenberger, line 1: Could you make the following changes here?
* Mark the reference implementation as work in progress with a disclaimer that it might not yet implement the latest changes of the standards.
* Add a Todo section with probably the following points:
  * Catch up with reference implementation
  * Extract crypto library for signature verification
  * Improve agent-js to expose the features we need here (to get rid of the copied code)

frederikrothenberger (approved):
Thanks a lot, Julia!
I'm already approving, as the readme changes are the only thing left. The subsequent changes we'll do in separate PRs.

PR #67, opened 2023-10-06
ICRC-25 ref impl: Refactor signer polling

frederikrothenberger:
This PR simplifies polling by making use of the
standard changes that all certified terminal states should (from the point of view of the signer) not be treated as an error, but simply returned to the relying party.

### DIFF
--- a/reference-implementations/icrc25/signer/src/agent.js
+++ b/reference-implementations/icrc25/signer/src/agent.js
@@ -23,60 +23,41 @@ export async function call(agent, canisterId, methodName, arg) {
   return await polling.pollForResponse(agent, Principal.from(canisterId), callResponse.requestId, polling.defaultStrategy())
 }
 
-export async function readState(agent, canisterId, requestId, strategy = polling.defaultStrategy()) {
-  return pollForResponse(agent, Principal.from(canisterId), requestId, strategy)
-}
-
-async function pollForResponse(agent, canisterId, requestId, strategy, request) {
-  const path = [new TextEncoder().encode('request_status'), requestId]
-  const currentRequest = request ?? (await agent.createReadStateRequest?.({ paths: [path] }))
-  const state = await agent.readState(canisterId, { paths: [path] }, undefined, currentRequest)
+export async function pollForCert(agent, canisterId, requestId) {
   if (agent.rootKey == null) throw new Error('Agent root key not initialized before polling')
-  const cert = await Certificate.create({
-    certificate: state.certificate,
-    rootKey: agent.rootKey,
-    canisterId: canisterId
-  })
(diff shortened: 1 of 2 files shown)


reference-implementations/icrc25/signer/src/agent.js
```diff
@@ -23,60 +23,41 @@ export async function call(agent, canisterId, methodName, arg) {
...
 
-  // eslint-disable-next-line default-case
-  switch (status) {
-    case RequestStatusResponseStatus.Replied: {
-      return { 
-        response: cert.lookup([...path, 'reply']),
-        certificate: state.certificate
-      }
+  for (;;) {
+    const request = await agent.createReadStateRequest({paths: [path]})
```
jsamol, line 31: From the original `pollForResponse` I gather that this could be created just once.

reference-implementations/icrc25/signer/src/agent.js
```diff
@@ -23,60 +23,41 @@ export async function call(agent, canisterId, methodName, arg) {
...
+      case RequestStatusResponseStatus.Replied:
+      case RequestStatusResponseStatus.Rejected:
+      case RequestStatusResponseStatus.Done:
+        return state.certificate
+
+      // continue polling
+      case RequestStatusResponseStatus.Received:
+      case RequestStatusResponseStatus.Unknown:
+      case RequestStatusResponseStatus.Processing:
+        continue;
```
jsamol, line 58: I'd keep using the polling strategy here (a default one used directly instead of being a parameter of this function should do fine) to introduce some sensible delays so the network has time to process the canister call, and optimize the amount of unnecessary network calls in result.

reference-implementations/icrc25/signer/src/agent.js
```diff
@@ -23,60 +23,41 @@ export async function call(agent, canisterId, methodName, arg) {
   return await polling.pollForResponse(agent, Principal.from(canisterId), callResponse.requestId, polling.defaultStrategy())
```
jsamol, line 23: `polling.defaultStrategy()` here is redundant

jsamol (approved):
A couple of minor things that could be improved, but nothing that prevents it from working correctly.

reference-implementations/icrc25/signer/src/agent.js
frederikrothenberger: It is not, for some reason `pollForResponse` does _not_ set a default if none is provided.

jsamol: Ah, you're right, my bad. For some reason I haven't noticed it was the library function being executed here, not the custom one some lines below.

PR #68, opened 2023-10-09
ICRC-25 ref impl: Replace signature verification code with library

frederikrothenberger:
This PR pulls in a library for IC signature verification.

### DIFF
--- a/reference-implementations/icrc25/relying-party/src/App.js
+++ b/reference-implementations/icrc25/relying-party/src/App.js
@@ -1,18 +1,14 @@
 /* global BigInt */
 import { BeaconEvent } from '@airgap/beacon-sdk'
 import { Certificate, requestIdOf } from '@dfinity/agent'
-import { Ed25519PublicKey, unwrapDER } from '@dfinity/identity'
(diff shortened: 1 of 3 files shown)


reference-implementations/icrc25/relying-party/src/App.js
```diff
@@ -69,81 +66,21 @@ function App() {
...
-
-    const publicKey = secp256k1PublicKey(derPublicKey) ?? ed25519PublicKey(derPublicKey) ?? p256PublicKey(derPublicKey)
-    if (publicKey === undefined) {
-      throw new Error('Public key not supported')
-    }
-
-    const signatureRaw = Buffer.from(response.result.signature, 'base64')
-    const dataRaw = Buffer.concat([
-      Buffer.from(new TextEncoder().encode('\x0Aic-wallet-challenge')), 
+    const derPublicKey = Uint8Array.from(Buffer.from(response.result.identities[0].publicKey, 'base64'))
```
jsamol, line 69: `Buffer` is a special type of`Uint8Array`, you shouldn't need to explicitly wrap it with `Uint8Array.from` 🤔

jsamol (approved):

PR #69, opened 2023-10-09
ICRC25 Error fixes

jsamol:
This PR fixes the error response definitions to adhere to the JSON-RPC 2.0 standard.

The errors are now identified by their 5-digit codes, where the first digit represents the version of the standard and the second digit defines the error's category. The categories defined for now are:
- General (x1xxx): errors whose cause can't be easily identified and categorized (e.g. Unknown error)
- Not supported (x2xxx): part of the message is not supported
- User action (x3xxx): the error occurred as a result of explicit user's action
- Network (x4xxx): network request failed

For example: 13002 suggests that it's an error defined in the version 1 of the standard and falls into the User action category.

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -13,7 +13,6 @@ This specification describes a communication protocol between dapps (decentraliz
 * signer: A service that manages a user's keys and can sign and perform canister calls on their behalf.
 * relying party: A service that wants to request calls on a specific canister.
 
-
 ## Transport Requirements
 
 This standard is agnostic to the transport channel used to send the messages, as long as it provides authenticity and integrity: this means that the communicating parties know the other participant and can be sure that the messages they receive are sent by the party they expect and that the messages have not been tampered with.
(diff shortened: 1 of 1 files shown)


topics/icrc_25_signer_interaction_standard.md
```diff
@@ -73,13 +74,28 @@ The purpose of the `permission` messages is to grant the relying party access to
 
 While processing the request from the relying party, the signer can cancel it at any time by sending an error in response.
 
-`errorType` (`text`): The reason behind the cancellation. Possible values:
-- `"VERSION_NOT_SUPPORTED`: The version of the standard is not supported by the signer.
-- `"NETWORK_NOT_SUPPORTED"`: The network on which the action was requested is not supported by the signer.
-- `"NOT_GRANTED"`: The signer has not granted permission to perform the action.
-- `"UNKNOWN"`: The reason is unknown.
+The error is an object comprising the `code`, `message` and optional `data` fields as described in the [JSON-RPC 2.0 Specification](https://www.jsonrpc.org/specification#error_object). In addition the the pre-defined errors, the following values are supported:
```
frederikrothenberger, line 77: I would make an `##Errors` category, and have the full table of all errors there. And then only list `possible errors: 11000, 12002, ...` on the message itself.

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -38,6 +38,7 @@ The purpose of the `permission` messages is to grant the relying party access to
 
 - `text`: A plain `string` value.
 - `blob`: A `string` value describing binary data encoded in base64.
+- `int` An integer value.
```
frederikrothenberger, line 41: Can you also lift out the `Types` into it's own section (i.e. `## Types`). I don't think it makes sense to define that multiple times.

frederikrothenberger (commented):
Thanks a lot, @jsamol! I like that you added additional structure to the errors.

But let me propose a slight change to the numbering system:
* Put the category first: This way, the parsing logic becomes much simpler to figure out the category (i.e. anything `<20000` is `general`).
* Allow 2 digits for the version, otherwise we're quite restricted on the number of iterations we can do.
* Put the version last. I.e. `category|err_no|version`. Because I suspect the relying party side of handling `32701` and `32705` to be the same in almost all cases 😉 
  * Alternatively, we can also leave the version off completely. Do you have a use-case in mind, where having error versions is important?

What do you think?

jsamol:
> - Allow 2 digits for the version, otherwise we're quite restricted on the number of iterations we can do.

Ah, good point 😄 

> Do you have a use-case in mind, where having error versions is important?

Simply that new versions could change the error definitions. Alternatively, we could have "unversioned" errors, mainly `Version not supported` that needs to be carried in unchanged form throughout any future versions, and implicitly "versioned" errors whose version will be determined based on the matching request.

frederikrothenberger (approved):
LGTM, thanks! 👍

PR #75, opened 2023-10-12
ICRC-25: Remove network selection

frederikrothenberger:
This PR simplifies the standard by removing the option to select networks. There is currently little benefit in having that feature, but it adds complexity to the standard.

If the network selection becomes useful later on, we can still add it either as an extension or by publishing a new version of the standard.


przydatek (approved):
LGTM, thanks!

PR #76, opened 2023-10-12
ICRC-29: Make method prefix consistent with ICRC-25

frederikrothenberger:
This PR makes the JSON-RPC 2.0 method prefix consistent with ICRC-25 messages.


przydatek (approved):
LGTM.

PR #79, opened 2023-10-13
ICRC-25: Split out identity information from `icrc25_request_permission` method

frederikrothenberger:
This PR splits the identitiy information part of the existing `icrc25_request_permission` method into another method called `icrc25_get_identities`. This makes the methods more focused.

The PR also adds clarifying information about how user interaction should be handled for the given methods. In particular, it allows signers to skip user interaction if the user previously approved the same request.

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -100,15 +100,16 @@ While processing the request from the relying party, the signer can cancel it at
     - If the request version is not supported by the signer, the signer sends a response with an error back to the relying party.
 3. The signer removes any unrecognized scopes from the list of requested scopes.
 4. Depending on the session state the signer either skips or displays the details of the to-be-established connection to the user:
-    - If there is no ongoing session with the relying party:
+    - If there is an active session with the relying party, skip to the next step, otherwise:
         - the signer presents the details of the to-be-established connection to the user. If the user has never interacted with this relying party before, the signer should display information explaining that the user is about to establish a connection with a new relying party.
+        - If the user approves the connection, the signer creates a new session for the relying party.
+            - Otherwise, the signer sends a response with an error back to the relying party and step 5 is skipped.
 
       > **Note:** The signer should maintain a list of relying parties that are trusted by the user. It is recommended that signers assist users when deciding to grant permissions to new relying parties, e.g. by maintaining a list of well-known relying parties and displaying additional information about the relying party, such as its name, logo, etc., or in the case of an unknown relying party, by displaying a warning.
-
-    - If there already is an active session with the relying party, the signer skips this step.
+      
 5. The signer displays the list of requested scopes to the user and asks the user to approve or reject the request. The user should also be allowed to approve only a subset of the requested scopes or add additional restrictions (see [optional scope restrictions](#optional-properties)).
     - If all requested scopes have already been granted, the signer may skip the user interaction and reply with the list of granted scopes immediately.
-    - If the user approves the request, the signer saves information about the granted permission scopes on the current session. If there is no active session, a new one must be created. Then the signer sends a successful response back to the relying party with the list of granted scopes.
+    - If the user approves the request, the signer saves information about the granted permission scopes on the current session. Then the signer sends a successful response back to the relying party with the list of granted scopes.
     - If the user rejects the request, the signer sends a response with an error back to the relying party.
 6. After receiving a response, the relying party may send additional messages depending on the granted scopes.
 
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -95,7 +95,7 @@ While processing the request from the relying party, the signer can cancel it at
 3. The signer removes any unrecognized scopes from the list of requested scopes.
 4. Depending on the session state the signer either skips or displays the details of the to-be-established connection to the user:
     - If there is no ongoing session with the relying party:
-        - the signer presents the details of the to-be-established connection to the user and asks the user to select identities that will be paired in response. If the user has never interacted with this relying party before, the signer should display information explaining that the user is about to establish a connection with a new relying party.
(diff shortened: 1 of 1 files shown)


topics/icrc_25_signer_interaction_standard.md
```diff
@@ -143,9 +144,109 @@ Request
...
+                "method": "icrc25_get_identities"
+            }
+        ]
+    }
+}
+```
+
+### `icrc25_get_identities`
+
+The purpose of the `icrc25_get_identities` message is for the relying party to receive information about the identities managed by the signer.
```
przydatek, line 177: It seems a bit confusing that the message and the scope use identical label `icrc25_get_identities`.  Is that intentional?

przydatek (commented):
I guess I'm missing something, but previously it was clear that the permission (e.g. for the scope `icrc25_canister_call`) is issued for the identities included in the response. With the current change, when we have two separate calls, it is not clear which identities are meant for which permissions.
Maybe add an example with a typical interplay of the scopes?  Or maybe we can discuss it next week?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -94,20 +89,22 @@ While processing the request from the relying party, the signer can cancel it at
...
-    assist users when deciding to grant permissions to new relying parties, e.g. by maintaining a list of well-known relying parties
-    and displaying additional information about the relying party, such as its name, logo, etc., or in the case of an unknown relying
-    party, by displaying a warning. 
-    - If the user approves the request, the signer saves information about the granted permission scopes and sends a successful response back to the relying party.
+3. The signer removes any unrecognized scopes from the list of requested scopes.
+4. Depending on the session state the signer either skips or displays the details of the to-be-established connection to the user:
+    - If there is no ongoing session with the relying party:
+        - the signer presents the details of the to-be-established connection to the user and asks the user to select identities that will be paired in response. If the user has never interacted with this relying party before, the signer should display information explaining that the user is about to establish a connection with a new relying party.
```
jsamol, line 98: > [...] asks the user to select identities that will be paired in response

I think this needs rephrasing, too, since we no longer return any identities in response. Maybe "asks the user to select identities that will be paired in the current session"? Or anything that implies that the identities selected in this step will be accessible throughout the session's lifetime and will be the only set of identities that the relying party will be allowed to execute actions on in the current session.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Yes, this is intentional. See the explanation in the `scopes` section:
> A scope is the permission to invoke specific JSON-RPC 2.0 method on the signer. A scope is identified by the `method` property which matches the `method` name of the JSON-RPC 2.0 call it relates to.

frederikrothenberger:
> I guess I'm missing something, but previously it was clear that the permission (e.g. for the scope `icrc25_canister_call`) is issued for the identities included in the response. With the current change, when we have two separate calls, it is not clear which identities are meant for which permissions. Maybe add an example with a typical interplay of the scopes? Or maybe we can discuss it next week?

The selection of identities and the permission scopes granted are orthogonal. This was also the case previously, but making it two separate calls makes this more explicit.

Both, identities and permission scopes are selected / granted for a specific session. And both of these things can change independently over the lifetime of that session.

I.e.:
* previously granted permission scopes can be revoked without changing the identities shared with the RP
* the list of identities shared with the RP can be adjusted during the session without changing permission scopes

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: I just removed all mention of identity selection from this use-case. Implementations are still free to prompt for identity selection out right, but IMHO it should not be mandated by the spec. WDYT?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -94,20 +95,22 @@ While processing the request from the relying party, the signer can cancel it at
...
+
+    - If there already is an active session with the relying party, the signer skips this step.
+5. The signer displays the list of requested scopes to the user and asks the user to approve or reject the request. The user should also be allowed to approve only a subset of the requested scopes or add additional restrictions (see [optional scope restrictions](#optional-properties)).
+    - If all requested scopes have already been granted, the signer may skip the user interaction and reply with the list of granted scopes immediately.
+    - If the user approves the request, the signer saves information about the granted permission scopes on the current session. If there is no active session, a new one must be created. Then the signer sends a successful response back to the relying party with the list of granted scopes.
```
przydatek, line 111: The paragraph states "If there is no active session, a new one must be created.", but it seems that at this point there should be an active session (created in step 4), no?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -94,20 +95,22 @@ While processing the request from the relying party, the signer can cancel it at
...
-    > **Note:** The signer should maintain a list of relying parties that are trusted by the user. It is recommended that signers
-    assist users when deciding to grant permissions to new relying parties, e.g. by maintaining a list of well-known relying parties
-    and displaying additional information about the relying party, such as its name, logo, etc., or in the case of an unknown relying
-    party, by displaying a warning. 
-    - If the user approves the request, the signer saves information about the granted permission scopes and sends a successful response back to the relying party.
+3. The signer removes any unrecognized scopes from the list of requested scopes.
+4. Depending on the session state the signer either skips or displays the details of the to-be-established connection to the user:
+    - If there is no ongoing session with the relying party:
```
przydatek, line 103: nit: maybe instead having this line, and then "- If there already is an active session with the relying party.." below, have something like: 
- If there is an active session with the relying party, skip to the next step, otherwise:"

przydatek (approved):
Thanks for the discussion and the adjustments. LGTM (modulo some minor suggestions).

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Ah, yes. This can be removed. Thanks.

PR #82, opened 2023-10-25
ICRC-25: Introduce method to query granted permissions

frederikrothenberger:
This PR introduces the method icrc25_granted_permissions method to query currently granted permission scopes.

It also renames icrc25_request_permission to icrc25_request_permissions so that it better reflects that multiple different permissions can be requested / granted.

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -27,7 +27,7 @@ The transport channel is not required to provide confidentiality.
 
 ## Sessions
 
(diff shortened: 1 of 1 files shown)


przydatek (approved):
LGTM, but I guess `icrc25_revoke_permission` should also eventually become `icrc25_revoke_permissions`, no?

frederikrothenberger:
> LGTM, but I guess `icrc25_revoke_permission` should also eventually become `icrc25_revoke_permissions`, no?

Yes, this was simply an oversight. Thanks, fixed.

PR #88, opened 2023-11-02
ICRC-21: Make method name more descriptive

frederikrothenberger:
In order to make the use-case of ICEC-21 clearer, the method icrc21_consent_message is renamed to
icrc21_canister_call_consent_message.

This rename is in anticipation of extensions to ICRC-21 that will introduce their own flavour of `consent_message` methods.


przydatek (approved):
LGTM, thanks!

PR #89, opened 2023-11-08
ICRC-25: Change use-case section titels to "message processing"

frederikrothenberger:
The section titles are renamed to better convey their content.

In addition, some broken ToC links were also fixed.


przydatek (approved):
LGTM, thanks!

PR #90, opened 2023-11-08
ICRC-25: Add delegation support to the `icrc25_managed_identities` message

frederikrothenberger:
This PR expands the message `icrc25_managed_identities` with an optional array of delegations, analogous to the delegations in canister call envelopes.
This enables signers that only have access to delegated keys to complete the challenge response mechanism.


frederikrothenberger:
@sea-snake: This PR enables your slide wallet to use the standard, right?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -338,6 +338,12 @@ The purpose of the `icrc25_managed_identities` message is for the relying party
 `identities`: A list of identities the user has selected to share with the relying party.
 - `publicKey` (`blob`): The DER-encoded public key associated with the identity, derived in accordance with one of [the signature algorithms supported by the IC](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures). The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
 - `signature` (`blob`): The signature produced by signing the concatenation of the domain separator `\x13ic-signer-challenge` (UTF-8 encoded) and the challenge with the private key associated with the identity.
+- `delegations` (optional): A list of delegations (as defined by the [IC interface specification, authentication section](https://internetcomputer.org/docs/current/references/ic-interface-spec/#authentication)):
```
przydatek, line 341: Maybe rename to `delegation_chain` or just `delegation`, as `delegations` indicates that there could be multiple delegations, and iiuc this is intended for a single delegation from `publicKey` to some other key, possibly via a "multi-step" delegation, right?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -356,7 +362,11 @@ While processing the request from the relying party, the signer can cancel it at
...
+    - If `delegations` is present and non-empty:
+      - The relying party validates that all delegations are not expired according to their respective `expiration` timestamp.
+      - The relying party validates all `signature` values of all delegations in the `delegations` list. For the first delegation in the array, this signature must be valid with respect to the public key corresponding to the `publicKey` field of the identity, all subsequent delegations are signed with the key corresponding to the public key contained in the preceding delegation. The key to validate the challenge signature with is the public key contained in the last delegation in the array.
+    - Otherwise the key to validate the challenge signature with is the public key contained in the `publicKey` field of the identity.
+    - The relying party retrieves the public key to verify the challenge from each identity according to the previous step, determines the `signature` scheme and verifies whether it was generated by signing the concatenation of the domain separator `\x13ic-signer-challenge` (UTF-8 encoded) and the `challenge` from the request with the private key associated with that public key.
```
przydatek, line 369: "according to the previous step" is somewhat misleading, as the method of getting the public key for the identity is actually described in the two previous steps.  Maybe restructure the description to merge the previous two steps into one, e.g. something like:
 - determine the public key for verifying the signature of the challenge for each identity: 
    - if there is no delegation (or delegation chain?), use `publicKey`
    - otherwise, verify the delegation chain, and use the public key from the last element of the chain

przydatek (commented):
Hi @frederikrothenberger, please see some minor suggestions below.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: > this is intended for a single delegation from publicKey to some other key, possibly via a "multi-step" delegation, right?

It is intended to have the exact same semantics as delegations have in the context of a canister call. Hence, why I'm referring to the interface spec here. There is no reason really to limit the delegations array to one entry only. In fact, this might only break with expectations of what developers have when interacting with the IC directly.
So, in short: `delegations` is IMHO appropriate.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Agreed, fixed.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: The interface spec refers to the field as `sender_delegation` (singular) so we will use the singular in the property name as well.

przydatek (approved):
LGTM, thanks!

PR #97, opened 2023-11-10
ICRC-29: Improve various wordings

frederikrothenberger:
This PR improves the wording of ICRC-29 in a few places. It clarifies the steps required to continue a session after the relying party has closed the signer window.


przydatek (approved):
LGTM, thanks!

PR #98, opened 2023-11-13
ICRC-25: List predefined errors explicitly

frederikrothenberger:
As requested in the working group, the list of predefined errors is mentioned explictly to make it less likely that devs miss those errors codes.


przydatek (approved):
LGTM, thanks!

PR #99, opened 2023-11-15
ICRC-21: Make error categories more intuitive

frederikrothenberger:
This PR simplifies & changes the `icrc21_error`
type to be more intuitive for developers.

In particular, it resolves the confusion around
`NotSupported` which could be interpreted in multiple ways.


przydatek (approved):
LGTM, thanks!

PR #102, opened 2023-11-21
ICRC-25: Add method to query supported standards

frederikrothenberger:
This PR introduces the `icrc25_supported_standards` method on the signer interface. This allows relying parties to easily query the extensions supported by the signer.

Currently in draft because the text already assumes ICRC-25 has been split into a base standard and extensions for functionality. Will be updated after the split.


plitzenberger (approved):
lgtm

PR #103, opened 2023-11-22
ICRC-31: extract icrc31_list_identities

plitzenberger:
This PR intends to extract the following standard extensions:

- **ICRC-31:** Get Principals
- **ICRC-32:** Sign Challenge
- **ICRC-3x:** Get Delegation

This is a collaborative change by @frederikrothenberger and @plitzenberger


topics/icrc_31_list_identities.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Summary
+
+The purpose of the `icrc31_list_identities` message is for the relying party to receive information about the identities managed by the signer.
+
+## Method
+
+**Name and Scope:** `icrc31_list_identities`
+
+**Prerequisite:** Active session with granted permission for method name/scope.
```
frederikrothenberger, line 25:
```suggestion
**Prerequisite:** Active session with granted permission scope `icrc31_list_identities` or `*`.
```

frederikrothenberger: Nit: I would make this as clear as possible to the reader what the exact permission needs to be.

topics/icrc_31_list_identities.md
```diff
@@ -0,0 +1,67 @@
...
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "version": "1",
+    "identities": [
+      {
+        "publicKey": "MFYwEAYHKoZIzj0CAQYFK4EEAAoDQgAEOTdHYwpFTr/oPXOfLQcteymk8AQE41VwPQ1W7Xpm0Zt1AY4+5aOnMAbAIjXEchxPuGbPWqPqwntXMPs3w4rOaA=="
```
frederikrothenberger, line 62: Here I would opt for transferring principals as the method is about _identities_ rather than public keys. Also, I think in most cases the relying party is only interested in the principal.
So my suggestion: `publicKey` -> `principal`

Also you might want to consider to just have an array of principals (textual representation so the type would be `string[]`) for most simplicity unless you expect the returned objects to be extended in the future. But allowing extensions on the returned identities also goes against the spirit of this restructuring that aims to extend the standards by introducing more method (rather than altering existing ones).

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+## Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`identities`: A list of identities the user has selected to share with the relying party.
+
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity, derived in accordance with one of [the signature algorithms supported by the IC](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures). The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
+- `signature` (`blob`): The signature produced by signing the concatenation of the domain separator `\x13ic-signer-challenge` (UTF-8 encoded) and the challenge with the private key associated with the identity.
```
frederikrothenberger, line 50: You dropped the delegation support here. Please add the optional `delegation` again (it is referenced in message processing).

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Summary
+
+The purpose of the `icrc32_sign_challenge` method is for the relying party to receive a cryptographic proof of ownership for the users identities.
+
+## Method
+
+**Name and Scope:** `icrc32_sign_challenge`
+
+**Prerequisite:** Active session with granted permission for method name/scope.
```
frederikrothenberger, line 21:
```suggestion
**Prerequisite:** Active session with granted permission scope `icrc32_sign_challenge` or `*`.
```

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`identities`: A list of identities the user has selected to share with the relying party.
+
+- `publicKey` (`blob`): The DER-encoded public key associated with the identity, derived in accordance with one of [the signature algorithms supported by the IC](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures). The public key can be used to [derive a self-authenticating principal](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal).
```
frederikrothenberger, line 49: Relating to my comment above, I would change the interface of the method to ask for a signature by `principal`: I.e. you select an identity to sign the challenge.

This also is more sound conceptually when we allow delegations: The returned challenge signature is _not_ created by the supplied public key / principal in that case, but a delegate of the identity associated with that public key / principal.

topics/icrc_3x_get_delegation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary
+
+The purpose of the `icrc3x_get_delegation` method is for the relying party to receive delegation identity.
+
+## Method
+
+**Name and Scope:** `icrc3x_get_delegation`
+
+**Prerequisite:** Active session with granted permission for method name/scope.
```
frederikrothenberger, line 21:
```suggestion
**Prerequisite:** Active session with granted permission scope `icrc3x_get_delegation` or `*`.
```

topics/icrc_3x_get_delegation.md
```diff
@@ -0,0 +1,65 @@
+# ICRC-3x: Get Delegation (ICRC-25 Extension)
+
+[![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31)
+
+<!-- TOC -->
+
+- [ICRC-32: Get Delegation](#icrc-3x-get-delegation)
```
frederikrothenberger, line 7:
```suggestion
- [ICRC-3x: Get Delegation](#icrc-3x-get-delegation)
```

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,110 @@
+# ICRC-32: Sign Challenge (ICRC-25 Extension)
+
+[![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31)
+
+<!-- TOC -->
+
+- [ICRC-32: Sign Challenge](#icrc-32-sign-challenge)
+  - [Summary](#summary)
+  - [Method](#method)
+  - [Request](#request-params)
```
frederikrothenberger, line 10: Request & Response are not valid references to existing headings.

topics/icrc_3x_get_delegation.md
```diff
@@ -0,0 +1,65 @@
...
+  - [Summary](#summary)
+  - [Method](#method)
+  - [Request](#request-params)
```
frederikrothenberger, line 10: Request & Response are not valid references to existing headings.

topics/icrc_31_list_identities.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+<!-- TOC -->
+
+- [ICRC-31: List Idenities](#icrc-31-list-identities)
+    - [Summary](#summary)
+    - [Method](#method)
+    - [Request](#request-2)
+    - [Response](#response-2)
+    - [Errors](#errors-2)
```
frederikrothenberger, line 13: This toc seems to be out of sync with the headings here.

frederikrothenberger (commented):
Thanks a lot! Nice work 🙂 

General comments:
* I would prefer the examples to have a heading so that they show up in the toc.
* I think it would be good to have a sentence about the base standard in the extensions (with a link to ICRC-25) in order to be more explicit than just the "(ICRC-25 Extension)" in the title.
* You dropped the `Errors` section. I think the extensions should also mention errors, either by referring to a base spec or keeping the list in the extension.
* I think the `get_delegation` ICRC draft should be split into a separate PR as it is clearly not as mature as the other two. That way we can have a more focused discussion about the new method in the separate PR (and not delay this one).

topics/icrc_31_list_identities.md
plitzenberger: Good point; I'm looking through my cross-chain glasses. That's why I've decided to move on with an object per identity. But I agree with you. Maybe we should be even more explicit and call this method `icrc_31_get_principal_ids?

frederikrothenberger: > Maybe we should be even more explicit and call this method `icrc_31_get_principal_ids?

I would suggest just `icrc_(list|get)_principals` which corresponds to the term that the [interface spec](https://internetcomputer.org/docs/current/references/ic-interface-spec/#principal) is using.

topics/icrc_31_get_principals.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Example RPC Response
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "version": "1",
+    "principalIds": [
```
frederikrothenberger, line 59:
```suggestion
    "principals": [
```

topics/icrc_31_get_principals.md
```diff
@@ -0,0 +1,70 @@
...
+    "version": "1"
+  }
+}
+```
+
+## Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`principals` (`text` array): A list of principalIds the user has selected to share with the relying party.
```
frederikrothenberger, line 49:
```suggestion
`principals` (`text` array): A list of principals the user has selected to share with the relying party.
```

topics/icrc_32_sign_challenge.md
plitzenberger: Updated; please check again 🙏

topics/icrc_31_get_principals.md
frederikrothenberger: This is not done?

topics/icrc_31_get_principals.md
```diff
@@ -0,0 +1,70 @@
...
+    "version": "1"
+  }
+}
+```
+
+## Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`principals` (`text` array): A list of principals the user has selected to share with the relying party.
```
frederikrothenberger, line 49:
```suggestion
`principals` (`text` array): A list of principals (textual representation) the user has selected to share with the relying party.
```

topics/icrc_31_get_principals.md
```diff
@@ -0,0 +1,70 @@
+# ICRC-31: Get Principals
+
+[![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31)
+[![Extention Badge](https://img.shields.io/badge/Extends-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
```
frederikrothenberger, line 4:
```suggestion
[![Extension Badge](https://img.shields.io/badge/Extends-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
```

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,126 @@
+# ICRC-32: Sign Challenge
+
+[![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31)
+[![Extention Badge](https://img.shields.io/badge/Extends-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
```
frederikrothenberger, line 4:
```suggestion
[![Extension Badge](https://img.shields.io/badge/Extends-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
```

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,126 @@
...
+      {
+        "publicKey": "MIIBIjANBgkqhkiG",
+        "signature": "MEUCIQDQ8Z3Z"
+      }
+    ]
+  }
+}
+```
+
+## Message Processing
```
frederikrothenberger, line 82: This section needs updating:
- wrong method name
- there is no identity selection anymore (as the RP chooses the identity) and only a single key is used to create s single signature
- step 5 needs to be updated to reflect that the response does no longer contain a list and that the RP needs to check that the signature corresponds to the identity specified in the request

topics/icrc_3x_get_delegation.md
```diff
@@ -0,0 +1,74 @@
+# ICRC-3x: Get Delegation
```
frederikrothenberger, line 1: I think we agreed to move this file to a different PR?

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,126 @@
...
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "version": "1",
+    "identities": [
+      {
+        "publicKey": "MIIBIjANBgkqhkiG",
```
frederikrothenberger, line 74: Maybe add a note that a valid example will be provided later.

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,126 @@
...
+    "challenge": "UjwgsORvEzp98TmB1cAIseNOoD9+GLyN/1DzJ5+jxZM="
+  }
+}
+```
+
+## Response
+
+`version` (`text`): The version of the standard used. It must match the `version` from the request.
+
+`identities`: A list of identities the user has selected to share with the relying party.
```
frederikrothenberger, line 53: I think we should rename this property to `signedChallenge` (or similar) and make it just an object (rather than a list) to better fit the new method.

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,126 @@
...
+    participant U as User
+
+    RP ->> S: Request identities
+    alt Version is not supported
+        S ->> RP: Error response: Version not supported (20101)
+    else Scope `icrc31_list_identities` not granted
+        S ->> RP: Error response: Permission not granted (30101)
+    else
+        opt If not selected before on the active session 
+            S ->> U: Ask to select identities to share with Relying Party
```
frederikrothenberger, line 115: Needs to be updated as well.

topics/icrc_3x_get_delegation.md
plitzenberger: I'm okay with both. Keeping it in this one is easier as we don't lose functionality. And as it is declared as a draft, it doesn't hurt to keep it here. What do you think?

frederikrothenberger: Ok, fair enough. 👍

topics/icrc_32_sign_challenge.md
```diff
@@ -0,0 +1,125 @@
...
+
+The purpose of the `icrc32_sign_challenge` method is for the relying party to receive a cryptographic proof of ownership for the users identities.
+
+## Method
+
+**Name and Scope:** `icrc32_sign_challenge`
+
+**Prerequisite:** Active session with granted permission scope `icrc32_sign_challenge` or `*`.
+
+## Request
```
frederikrothenberger, line 28: I just noticed, this section is missing the `principal` parameter.

PR #116, opened 2023-12-15
Feature/ICRC-34-update-number-and-references

plitzenberger:
This PR intends to update the standard with the issued id.


topics/icrc_34_get_delegation.md
```diff
@@ -1,10 +1,12 @@
-# ICRC-3x: Get Delegation
+# ICRC-34: Get Delegation
 
 [![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31)
-[![Extension Badge](https://img.shields.io/badge/Extends-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
+[![Extension Badge](https://img.shields.io/badge/EXTENDS-ICRC--25-ffcc222.svg)](./icrc_25_signer_interaction_standard.md)
+[![Standard Issue](https://img.shields.io/badge/ISSUE-ICRC--34-blue?logo=github)](https://github.com/dfinity/wg-identity-authentication/issues/115)
```
frederikrothenberger, line 5: I'm not sure we want to really link this?
From the issue to the standard, absolutely. But having the issue link in the standard is not super meaningful. Especially when we finalize the standard, the issue will not have any significance any more...

Although I have no strong opinion on this. It might be helpful during the standardization process maybe...

frederikrothenberger (approved):
Thanks!

topics/icrc_34_get_delegation.md
plitzenberger: Yea, my reasoning was that it's easy to find the Issue by having this back reference. So that we remember to update it accordingly.

PR #122, opened 2024-01-10
Explain differences between signer identities and delegation identities

frederikrothenberger:
This PR adds a small section about the differences between delegation identities and signer identities.
Thanks @seniorjoinu for pointing out that this information was missing!


przydatek (approved):
LGTM, thanks!

PR #126, opened 2024-01-18
ICRC-33: Clarify blind signing and add stronger warning

frederikrothenberger:
This PR makes it clear that supporting blind signing is dangerous and not required. In addition it now mandates that if signers do support blind signing, that the users need to opt in explicitly.

A new error to abort on missing consent message is introduced.


frederikrothenberger:
Thanks for spotting all the typos! 🙂

Issue #131, opened 2024-01-25
ICRC-25: Specifying user interactions


frederikrothenberger:
So far, ICRC-25 and it's extensions have in some cases left open when to interact with the user.
It was brought up by @sea-snake in the last working group session that this makes it confusing for relying parties to provide a consistent UX for the signer interactions.

This issue should be used to discuss that point.

---

My personal opinion on the topic:

I think it would be good to specify for each signer method how they interact with the user:

| Method | Requires approval | Notes |
| --- | --- | --- |
| icrc25_request_permissions | yes | It's up to the signer _how_ to describe the requested permissions, but they need to show a screen that the user has to confirm.<br><br>Note: There might exist a use-case for institutional grade signers, where a user that has access to the signer might not have the authority to grant certain permissions. If the set of approvable permissions is empty, the approval screen can be skipped and a response can be sent immediately.<br><br>TLDR: Requires approval if the set of possible new permissions is larger than the currently approved one. |
| icrc25_granted_permissions | No | This is simply querying session state. |
| icrc25_revoke_permissions | No | Reducing permission does not require approval. It is up to the signer to notify the user. |
| icrc25_supported_standards | No | This is simply querying signer capabilities. |
| icrc27_get_icrc1_accounts | Yes | Prompts the user to select / reconfirm the accounts. |
| icrc31_get_principals | Yes | Prompts the user to select / reconfirm the connected signer identities. |
| icrc32_sign_challenge | Yes | Prompts the user to confirm the challenge signature |
| icrc34_get_delegation | No | The user has already approved that the dapp may retrieve a scoped delegation when granting the permission. The actual exchange / signing of the delegation is a technical detail. |
| icrc49_call_canister | Yes | Prompts the user to approve a transaction |

I think for an MVP framework of standards, this should be good enough.

More advanced use-cases, such as
* getting notified upon changes (i.e. account name change)
* querying already approved data non-interactively
* ...

should IMHO be solved within their own standards and at a later stage. I think it is important to first validate basic MVP use-cases with the tools given by the methods above (which arguably already go beyond an MVP) and then later extend ICRC-25 further where there is a clear need for improved UX / not supported use-cases.

---

@sea-snake, @plitzenberger, @kristoferlund, @dostro: Please share your opinion on this.

plitzenberger:
@frederikrothenberger I like the proposal! My two cents on:

> getting notified upon changes (i.e. account name change)

[...]

PR #132, opened 2024-01-26
Update ICRC-27 based on feedback

sea-snake:
After the last ICRC-27 meeting there was some feedback:

- Return all ICRC-1 accounts at once
- Clearify the standard in relation to ICRC-31 + ledger index canister
- How to trigger re-selecting accounts instead of returning previously selected

In this PR I've tried to update the ICRC-27 to address above points.

~~Regarding interactivity, I've opted for a required method argument over having separate methods for the following reasons:~~

- ~~With two methods, you would need to ask permission for both~~
- ~~With two methods, you would need to unnecessarily implement the same method signature twice~~
- ~~Some signers might not support interactive account selection, this is more clear with a single method~~
- ~~For the relying party the single method would behave same as long the signer supports the non interactive argument as required~~

Edit regarding interactivity, I've opted to make the get accounts method always interactive (prompt user to select accounts) as discussed in https://github.com/dfinity/wg-identity-authentication/issues/131.


README.md
```diff
@@ -4,22 +4,23 @@
...
-| [CAIP-2](https://github.com/icvc/icp-namespace/pull/1)<br>[CAIP-10](https://github.com/icvc/icp-namespace/pull/2) | CAIP Specifications                                                                                                                                               | [Quint Daenen](https://github.com/q-uint)                         | [![Status Badge](https://img.shields.io/badge/STATUS-DRAFT-ffcc00.svg)](https://github.com/orgs/dfinity/projects/31) | [Issue 25](https://github.com/dfinity/wg-identity-authentication/issues/25), [CAIP Specifications](https://github.com/ChainAgnostic/CAIPs) |
+| [CAIP-2](https://github.com/icvc/icp-namespace/pull/1)<br>[CAIP-10](https://github.com/icvc/icp-namespace/pull/2) | CAIP Specifications
```
frederikrothenberger, line 23: You should probably not remove this. At least not without talking to Quint first. 😉

topics/icrc_27_get_icrc_1_accounts.md
```diff
@@ -0,0 +1,204 @@
+# ICRC-27: Get ICRC-1 Accounts
```
frederikrothenberger, line 1: This PR only adds a file. You should also delete the other existing file `icrc_27_get_icrc_1_subbaccounts.md`.

frederikrothenberger (commented):
Very nice work, only some clean-up left.

topics/icrc_27_get_icrc_1_accounts.md
sea-snake: Fixed, PR now correctly shows file as renamed.

README.md
sea-snake: Fixed also double checked that nothing was removed and correctly updated the table column alignment

PR #133, opened 2024-01-26
ICRC-33/49: Rename ICRC-33 to ICRC-49

kristoferlund:
Renaming occurences of ICRC-33 to ICRC-49.


kristoferlund:
@frederikrothenberger came across some mentions of ICRC-33.

PR #134, opened 2024-01-28
Fix JSON typing in example to follow JSON RPC

sea-snake:
Fix JSON typing in example to follow JSON RPC standard (should be integer)


frederikrothenberger (approved):
LGTM, thanks!

PR #135, opened 2024-01-29
Fix delegations data structure

sea-snake:
Fix delegations data structure, make sure that delegation field is explicitly defined as defined in the IC spec. Also renames top level optional delegation field to delegations to indicate it's a list of delegations.

One question here, what encoding do we use for blobs here? I assume base64, in that case we should probably specify that. Right now the different icrc25 extensions use a mix of blob, text and other types :/


frederikrothenberger:
@sea-snake : I took the liberty of implementing the review feedback myself in order to unblock the alignment with recent changes in ICRC-25.
Feel free to reach out, create another PR, etc if I missed something.

PR #137, opened 2024-02-05
ICRC-21: Fix error in consent message validation

frederikrothenberger:
This PR fixes an error in the consent message validation that was still referring to an old version of the target canister interface.

In addition, the diagrams are improved to make it clearer that both the consent message request and response are required for validation in the (cold) signer commponent.


przydatek (approved):
LGTM, thanks!

Issue #128, opened 2024-01-24
ICRC-25: Version negotiation


frederikrothenberger:
Currently ICRC-25 specifies `version` fields in both requests and responses. If the signer does not have the exact matching version, it is supposed to send an error back to the relying party.

As it currently stands, this makes it hard for relying party to find out which version is actually supported by the signer.

Let's use this issue to discuss how we want to proceed with the `version` field.

Options (as I currently see them):

| Option | Description | Versioning Mechanism |
| --- | --- | --- |
| 1 | Drop the `version` field | New versions will have a new ICRC number with a new method, discoverable using `icrc25_supported_standards` |
| 2 | Allow the relying party to send an array of versions | The relying party sends the list of all versions it supports, the signer picks the highest from the list it supports (or sends `version_not_supported` if there is none) |
| 3 | Allow the relying party to send the highest supported version | The relying party sends only the highest supported version. It is expected to support all lower versions. The signer picks the highest version it supports that is equal or lower to the relying party highest version (or sends `version_not_supported` if there is none)

Personally, I'm in favor of option 1 because it simplifies our standards. Using the `icrc25_supported_standards` method we can easily provide improved versions of existing methods. This also aligns better with the standardization process that does not really allow to publish new versions of existing standards. On top of that, given every method has its own standard, this allows signers to drop support for the old version by simply not including it in the list of supported standards.

@plitzenberger, @sea-snake: What is your opinion on that?

plitzenberger:
@frederikrothenberger I'm voting for `option 1` and introducing versioned methods.

frederikrothenberger:
> introducing versioned methods

Could you elaborate on that? I.e. you would do `icrc31_get_principals_v2`? Although `icrc31_get_principals_v2` would be specified in some other ICRC? We could also simply offer `icrcXX_get_principals`, where XX is > 31 indicating that it is a later standard?

sea-snake:
My vote would be for option 1, it simplifies the method signature by keeping it constant (no versions = no changes) while allowing for new methods (for e.g. new versions) with new icrc standards analogues to how icrc canister standards are currently evolving (new or extending standards over versioning within a standard).

frederikrothenberger:
As discussed in the working group meeting on February 06, 2024 the `version` field will be dropped. If there is a need for an improved version of an existing method, a new ICRC standard should be made to describe the method.

Given ICRC numbers are part of the method name, we can then name the method the same, except for that number.

[...]

PR #140, opened 2024-02-08
ICRC-25: Drop version and simplify errors

frederikrothenberger:
This PR implements the decision taken in #128 for ICRC-25. This means that all `version` fields are dropped. This has an effect on errors as they used to include the version in the code. This is now simplified to just indicate the broader category with the first digit and the rest being incremental numbers for each error. The "unknown error" is dropped in favor of a "generic error" with a description to avoid unhelpful errors.

By dropping the `version` field, some requests have no `params` anymore. To not confuse readers by writing "Request: none", the heading is changed to "Request Params", to better align with the JSON RPC spec and to make it clear that there still is a request being sent.
For consistency, all "Response" headings are changed to "Result".

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -16,34 +16,33 @@
   * [Methods](#methods)
     * [`icrc25_request_permissions`](#icrc25_request_permissions)
       * [Prerequisites](#prerequisites)
-      * [Request](#request)
-      * [Response](#response)
+      * [Request Params](#request-params)
+      * [Result](#result)
       * [Errors](#errors)
       * [Message Processing](#message-processing)
       * [Example](#example)
     * [`icrc25_granted_permissions`](#icrc25_granted_permissions)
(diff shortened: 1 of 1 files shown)


topics/icrc_25_signer_interaction_standard.md
```diff
@@ -297,10 +267,7 @@ Request
 {
     "id": 1,
     "jsonrpc": "2.0",
-    "method": "icrc25_granted_permissions",
-    "params": {
-        "version": "1"
-    }
+    "method": "icrc25_granted_permissions"
```
przydatek, line 270: This example (and other examples below) still uses "Response" (line 274) rather than "Result".

przydatek (approved):
LGTM, thanks!

topics/icrc_25_signer_interaction_standard.md
przydatek: Ah, I see.  Sorry for the noise.

PR #142, opened 2024-02-08
ICRC-21: Point standard link to ICRC repo

frederikrothenberger:
When finalized, the standards need to be available in the ICRC repo. To not upend how this working group operates, we will continue in this repository while the standards are in draft. However, we prepare the drafts such that the transition is easy once finalize. To that end, we already point to the location where the standard will end up eventually when querying supported standards.


przydatek (approved):
LGTM, thanks.

PR #143, opened 2024-02-08
ICRC-31: Drop version and specify user interaction

frederikrothenberger:
In alignment with #128 and #131:
* the `version` is dropped
  * this includes the changes to the errors made in #142
* user interaction is made explicit


przydatek (approved):
LGTM, thanks!

PR #146, opened 2024-02-08
ICRC-32: Add examples

frederikrothenberger:
Add examples of actual challenge signature with and without delegation.


przydatek (approved):
LGTM, thanks!

PR #152, opened 2024-02-12
ICRC-27: Drop version and align errors

frederikrothenberger:
This PR drop `version` in alignment with #128. The errors are adjusted to that change as well.


przydatek (approved):
LGTM, thanks!

PR #154, opened 2024-02-12
ICRC-32: Simplify result and avoid duplication of delegation field name

frederikrothenberger:
This simplifies the result by flattening the outermost object. This object was initially created because the method used to be able to provide multiple signatures (which is no longer the case to avoid complexity).

In addition, the outermost `delegation` field is renamed to `signer_delegation`. This is consistent with `sender_delegation` in the canister call envelope and avoids having multiple `delegation` fields on different hierarchy levels.

### DIFF
--- a/topics/icrc_32_sign_challenge.md
+++ b/topics/icrc_32_sign_challenge.md
@@ -85,11 +85,11 @@ An ICRC-25 compliant signer must implement the [icrc25_supported_standards](./ic
 
 ## Result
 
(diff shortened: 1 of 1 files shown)


przydatek (approved):
LGTM, thanks!
nits: the PR description seems to have some typos: "simplifys"-> "simplifies", "cause" -> "case"

frederikrothenberger:
> LGTM, thanks! nits: the PR description seems to have some typos: "simplifys"-> "simplifies", "cause" -> "case"

Thanks 🙈

PR #157, opened 2024-03-14
ICRC-21: Improve user preferences to be able to handle line displays

frederikrothenberger:
This PR refactors how the user preferences are handled in the ICRC-21 standard:
* There is now a single `icrc21_consent_message_spec` present in both the request and the response. This makes it easier for signers to compare the requested spec to the actual spec.
* The `icrc21_consent_message_spec` can now be expanded without affecting other types (previously the language preference was duplicated in the `icrc21_consent_preferences` and `icrc21_consent_info`).
* The `icrc21_consent_message_spec` is expanded to handle line displays. This is required for devices such as thee Ledger hardware wallets to process ICRC-21 consent messages.

The candid interface is split into a separate file to make it easier to use for developers.


topics/ICRC-21/ICRC-21.did
```diff
@@ -0,0 +1,142 @@
...
+    // │  ┌───────────────────────────┐  │
+    // │  │    <consent_message>      │  │
+    // │  └───────────────────────────┘  │
+    // │  ┌───────────┐   ┌───────────┐  │
+    // │  │  Reject   │   │  Approve  │  │
+    // │  └───────────┘   └───────────┘  │
+    // └─────────────────────────────────┘
+    consent_message: icrc21_consent_message;
+    // Actual consent message spec used to generate the message.
+    // Clients must reject the consent message if the spec does not match the supplied consent_message.
```
przydatek, line 77: nit: this implies that the client must recognise  e.g. whether the message is in the specified language, which might be difficult in practice.  Or is the rejection here meant to be "best effort"?

topics/ICRC-21/ICRC-21.did
```diff
@@ -0,0 +1,142 @@
+// Canister interface specification for ICRC-21.
+// See https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md
+
+type icrc21_consent_message_spec = record {
+    // BCP-47 language tag. See https://www.rfc-editor.org/rfc/bcp/bcp47.txt
+    language: text;
+    // Information about the device responsible for presenting the consent message to the user.
+    device_spec: variant {
+        // A generic display able to handle large documents and do line wrapping and pagination / scrolling.
+        // Text must be Markdown formatted, no external resources (e.g. images) are allowed.
```
przydatek, line 10: nit: since we already are having variants, maybe it would make sense to explicitly distinguish also between `GenericDisplayText` and `GenericDisplayMarkdown`, so that FE-developers don't have to guess (or rely on some library's guess) whether the message is in Markdown or not?

przydatek (approved):
LGTM, with two minor questions.  Thanks!

topics/ICRC-21/ICRC-21.did
frederikrothenberger: The idea is to restrict the flexibility. I.e. `GenericDisplayText` is _always_ Markdown and `LineDisplayMessage` must be plain text.

That way developers don't have to deal with 4 options.

topics/ICRC-21/ICRC-21.did
frederikrothenberger: @przydatek: I have now bundled non-enforced `metadata` from the `consent_message_spec`. Now the `device_spec` is no longer included in the response and hence cannot cause a conflict with the `icrc21_consent_message` variant, thus sidestepping the issue.

Please take another look.

topics/ICRC-21/ICRC-21.did
```diff
@@ -0,0 +1,147 @@
...
+            // Lines must not exceed the number of characters per line specified in the icrc21_consent_message_spec.
+            lines: vec text;
+        };
+    };
+};
+
+type icrc21_consent_info = record {
+    // Consent message describing in a human-readable format what the call will do.
+    //
+    // The message should adhere as close as possible to the user_preferences specified in the consent_message_spec
```
frederikrothenberger, line 60: WG feedback: Clarify on flexibility

frederikrothenberger:
@sea-snake: I made the `device_spec` optional in the `icrc21_consent_message_spec` to address the points raised in working group meeting that we had earlier today. I also clarified the comments to be more explicit about fallbacks.

przydatek (approved):
LGTM, thanks!

sea-snake (approved):
Looks good to me :D

PR #160, opened 2024-03-22
Remove id requirements from ICRC-39

### LINKED ISSUE #136: ICRC-39 batch sequential id requirement
I read the following in the ICRC-39 spec while implementing it:

> Order all requests in the batch by their id in ascending order.

When I read the JSON RPC spec, this requirement does not exist, it even defines that the responses could be out of order and ids should be used:

> The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object.

Also since ids could also be strings besides numbers in the JSON RPC spec and the signer will receive the batch in correct order already since it's a list of requests, I would opt to not have this ascending order requirement on the ids.

sea-snake:
Remove id requirements from ICRC-39 and clarify that the array positions of the received requests must be equal to the corresponding responses.

Closes #136.

### DIFF
--- a/topics/icrc_39_batch_calling.md
+++ b/topics/icrc_39_batch_calling.md
@@ -17,14 +17,13 @@
 
 JSON-RPC defines a [batch call](https://www.jsonrpc.org/specification#batch) as a JSON array of requests. By supporting this standard a signer declares that all methods offered by the signer may also be invoked as part of a batch.
 
-If a signer receives a batch call, it must process each request sequentially in order of the id and reply with a batch response. If any call results in an error response the signer must not process any further calls in the batch but add an error for any not executed call in the batch.
+If a signer receives a batch call, it must process each request sequentially in order of array position and reply with a batch response. If any call results in an error response the signer must not process any further calls in the batch but add an error for any not executed call in the batch. The signer **MUST** return exactly one response for every request in the same order and with the same ids as the received requests.
 
 ## Batch Processing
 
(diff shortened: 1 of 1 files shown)


topics/icrc_39_batch_calling.md
```diff
@@ -17,14 +17,13 @@
...
 
-If a signer receives a batch call, it must process each request sequentially in order of the id and reply with a batch response. If any call results in an error response the signer must not process any further calls in the batch but add an error for any not executed call in the batch.
+If a signer receives a batch call, it must process each request sequentially in order of array position and reply with a batch response. If any call results in an error response the signer must not process any further calls in the batch but add an error for any not executed call in the batch. The signer **MUST** return exactly one response for every request in the same order and with the same ids as the received requests.
 
 ## Batch Processing
 
 Batch call must be processed by signers as follows:
 
-1. Order all requests in the batch by their id in ascending order.
-2. Sequentially, for each request in the batch
```
frederikrothenberger, line 27: Please change (top-level) step `3.` -> `2.`

frederikrothenberger (commented):
Thank you @sea-snake! One minor comment.

topics/icrc_39_batch_calling.md
sea-snake: Updated

frederikrothenberger (approved):
Thanks!

PR #170, opened 2024-05-05
Combine ICRC-27 and ICRC-31 into a simplified single standard.

sea-snake:
As discussed in the last WG meeting, the ICRC-27 and ICRC-31 standards have been combined in this PR into a single simplified standard. 

Limitations defined in the earlier ICRC-27 standards have been removed and a section has been added regarding handling (un)supported token standards as relying party. 

Additionally, user interaction requirements have been removed from the diagram and replaced with a recommendation regarding consistent signer behavior.

PR #173, opened 2024-05-07
ICRC-25: Remove wildcard permission and support stateless signers

frederikrothenberger:
This PR updates ICRC-25 with the following changes:
* Remove wildcard scope
* Make sessions optional
* Make user-interaction optional on `icrc25_request_permissions` when prompting the user on every invocation of extension methods
* Fix typo in sequence diagram of `icrc25_supported_standards`
* Adjust examples to #170

### DIFF
--- a/topics/icrc_25_signer_interaction_standard.md
+++ b/topics/icrc_25_signer_interaction_standard.md
@@ -11,7 +11,6 @@
   * [Sessions](#sessions)
   * [Scopes](#scopes)
     * [Scope Objects](#scope-objects)
-    * [Scopes Defined by this Standard](#scopes-defined-by-this-standard)
   * [Extensions](#extensions)
   * [Methods](#methods)
     * [`icrc25_request_permissions`](#icrc25_request_permissions)
@@ -69,11 +68,13 @@ The transport channel is not required to provide confidentiality.
 
 ## Sessions
 
-ICRC-25 uses sessions to determine the lifetime of granted permission [scopes](#scopes). Permission scopes (see [`icrc25_request_permissions` message](#icrc25_request_permissions)) are granted for the duration of a single session only. 
+ICRC-25 may use sessions to store user-choices and determine the lifetime of granted permission [scopes](#scopes). Permission scopes (see [`icrc25_request_permissions` message](#icrc25_request_permissions)) are granted for the duration of a single session only. 
 
 A session is established when the first permission request is granted. A session can be revoked by the relying party at any time by sending a [`icrc25_revoke_permissions` message](#icrc25_revoke_permissions) to the signer. The signer can also terminate the session at any time and should offer the user a method to do so.
 
 A session must be terminated automatically after a certain period of inactivity. The session might be extended automatically if the interaction between the relying party and the signer is still _actively_ ongoing when the default session timeout is reached. There must be a maximum session duration (regardless of activity).
+
+Signers that do not maintain sessions (i.e. stateless signers) MUST prompt the user to approve each request individually for each method defined in an ICRC-25 extension.
 
 ## Scopes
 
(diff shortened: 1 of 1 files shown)


topics/icrc_25_signer_interaction_standard.md
```diff
@@ -88,40 +89,18 @@ Scopes are represented in JSON-RPC 2.0 messages as JSON objects with the followi
...
-```
-
-Extensions to this standard may define additional scopes.
-
 ## Extensions
 
 This standard is the signer interaction _base_ standard. As such it intentionally excludes all methods that could be handled by an extension, for example:
 
-- Getting principals: [ICRC-31](./icrc_31_get_principals.md)
+- Getting principals: [ICRC-27](./icrc_27_get_accounts.md)
```
dostro, line 96: Getting accounts

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -88,40 +89,18 @@ Scopes are represented in JSON-RPC 2.0 messages as JSON objects with the followi
...
 - Proving ownership of principals: [ICRC-32](./icrc_32_sign_challenge.md)
```
dostro, line 97: Proving ownership of accounts(?)
We're not prioritizing signing a challenge right now so it's unclear what this will be. Should we remove for now?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -69,12 +68,14 @@ The transport channel is not required to provide confidentiality.
...
+ICRC-25 may use sessions to store user-choices and determine the lifetime of granted permission [scopes](#scopes). Permission scopes (see [`icrc25_request_permissions` message](#icrc25_request_permissions)) are granted for the duration of a single session only. 
 
 A session is established when the first permission request is granted. A session can be revoked by the relying party at any time by sending a [`icrc25_revoke_permissions` message](#icrc25_revoke_permissions) to the signer. The signer can also terminate the session at any time and should offer the user a method to do so.
 
 A session must be terminated automatically after a certain period of inactivity. The session might be extended automatically if the interaction between the relying party and the signer is still _actively_ ongoing when the default session timeout is reached. There must be a maximum session duration (regardless of activity).
 
+Signers that do not maintain sessions (i.e. stateless signers) MUST prompt the user to approve each request individually for each method defined in an ICRC-25 extension.
```
dostro, line 77: Let's see how MUST goes in practice. Imagine a situation with sign_challenge: Why would a user need to approve a simple challenge signature simply to prove they own the account instead of having the signer auto-sign? There doesn't seem to be any reason against that.

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -151,19 +130,21 @@ While processing the request from the relying party, the signer can cancel it at
 
 1. The relying party sends a `icrc25_request_permissions` message to the signer.
 2. The signer removes any unrecognized scopes from the array of requested scopes.
-3. Depending on the session state the signer either skips or displays the details of the to-be-established connection to the user:
+3. Depending on whether the signer supports sessions:
+   - If the signer does _not_ support sessions, it may send a response immediately _without_ prompting the user. Scopes should be granted subject to a static signer policy. Skip to step 6.
+     > **Note:** Stateless signers MUST prompt the user to approve each request individually for each method defined in an ICRC-25 extension.
+   - Otherwise, continue with step 4.
+   > **Note:** It is recommended that signers assist users when granting permissions to relying parties, e.g. by maintaining a list of well-known relying parties and displaying additional information about the relying party, such as its name, logo, etc., or in the case of an unknown relying party, by displaying a warning.
```
dostro, line 137: Why are we adding a recommendation for such a substantial centralized effort?

topics/icrc_25_signer_interaction_standard.md
```diff
@@ -213,7 +198,7 @@ Response
     "result": {
         "scopes": [
             {
-                "method": "icrc31_get_principals"
+                "method": "icrc27_get_accounts"
             },
             {
                 "method": "icrc49_call_canister",
```
dostro, line 204: Response for granted permissions shouldn't contain any information about the sender

dostro: If nothing else, it's a loophole in permissions requests:
- Dapp requests accounts -> user denies
- Dapp requests calling canister -> user accepts and sends a principal anyway, bypassing the denied request for accounts. Also, how would the user select their principal(s) for this permission?

dostro (changes requested):

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Thanks!

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: Yes, let's remove this for now.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: I mean, I think it is a good recommendation (if the resources to implement it are available). The list must not necessarily be maintained in a centralized way...

But I see that this does not need to be in the standard necessarily. It is probably better suited in some additional document about signer implementation recommendations / best practices.

Removed.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: I think the original intention was to allow signers to share multiple identities but only allow transactions with a subset of these. Of course, a sensible signer implementation would then tie the restriction to the user-choice of sharing the identity in the first place (and not include it, if the identity wasn't shared).

We should probably have a discussion about the additional scope restrictions of ICRC-49.
I removed it for now.

topics/icrc_25_signer_interaction_standard.md
frederikrothenberger: > There doesn't seem to be any reason against that.

That depends heavily on what the challenge signatures are used for. Let's discuss this in today's session.

frederikrothenberger: Removed, as discussed in the last working group.

dostro (approved):

PR #176, opened 2024-05-08
ICRC-29: broadcast ready message

peterpeterparker:
# Motivation

The `icrc29_ready` cannot be scoped to a particular origin because the origin is unknown at the time the notification is emitted. Therefore, I think it is worth highlighting in the specification that this is a broadcast.

# Notes

Limitation, or approach, is  notably similar to Internet Identity specification for `authorize-ready`.

### DIFF
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -36,7 +36,7 @@ using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/
 ## Authentication
 
 * The relying party must authenticate the signer by the `origin` of the window that it opens.
-* The signer broadcasts the "ready" message cannot authenticate the relying party by the `origin` property of the `message` event that it received.
+* The signer broadcasts the "ready" message to its parent.
 
 ## Sending Messages
 
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -36,12 +36,11 @@ using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/
 ## Authentication
 
 * The relying party must authenticate the signer by the `origin` of the window that it opens.
-* The singer must authenticate the relying party by the `origin` property of the `message` event that it received.
+* The signer broadcasts the "ready" message cannot authenticate the relying party by the `origin` property of the `message` event that it received.
 
 ## Sending Messages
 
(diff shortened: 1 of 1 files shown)


topics/icrc_29_window_post_message_transport.md
```diff
@@ -36,12 +36,11 @@ using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/
 ## Authentication
 
 * The relying party must authenticate the signer by the `origin` of the window that it opens.
-* The signer must authenticate the relying party by the `origin` property of the `message` event that it received.
```
frederikrothenberger, line 39: While the `ready` message is indeed a broadcast, it is still true that the signer authenticates the relying party by message event origin.
It just means that the ready message has to be broadcast _before_ the origin is known.
I would leave this sentence as is, given this is the "Authentication" section.

topics/icrc_29_window_post_message_transport.md
```diff
@@ -36,12 +36,11 @@ using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/
...
+* The signer broadcasts the "ready" message to its parent.
 
 ## Sending Messages
 
-Messages are sent by calling `window.postMessage` on the signer window, or the `window.opener` respectively.
-When sending messages, the `targetOrigin` parameter must be set to the origin of the signer or relying party window.
+The "ready" message is sent by the signer to the relying party using either `parent.postMessage` or `window.opener.postMessage`, specifically targeting the potential parent that opened the signer window. This message is broadcast without being scoped to any specific origin. This approach is necessary because it initiates communication, and until data exchange begins, the origin of the relying party remains unknown.
```
frederikrothenberger, line 43: Thanks for this clarification. I think we should keep the two sentences that you replaced and make them specific to sending responses to specific requests.

What do you think?

topics/icrc_29_window_post_message_transport.md
peterpeterparker: Generally speaking yes but, this document specifies ICRC-29. In this particular case the signer cannot authenticate the origin. Therefore I personnaly would adapt those chapter "Authentication" and "Sending Messages" to the particular spec it relates but, of course, this PR is just a suggestion.

peterpeterparker:
Discussed offline. I understood this specification as the specification of the "ready" message while it actually covers the general data exchange flow. There I close this PR and let @frederikrothenberger provides adjustments to the structure of the document if necessary in another PR.

PR #177, opened 2024-05-14
Switch supported standards method to ICRC-10

frederikrothenberger:
Change the icrc21_supported_standards query to
icrc10_supported_standards to be in line with the changes made in #171.

### DIFF
--- a/reference-implementations/ICRC-21/src/lib.rs
+++ b/reference-implementations/ICRC-21/src/lib.rs
@@ -15,11 +15,17 @@ fn greet(name: String) -> String {
 }
 
 #[ic_cdk::query]
-fn icrc21_supported_standards() -> Vec<Icrc21SupportedStandard> {
-    vec![Icrc21SupportedStandard {
-        url: "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md".to_string(),
-        name: "ICRC-21".to_string(),
-    }]
+fn icrc10_supported_standards() -> Vec<Icrc21SupportedStandard> {
+    vec![
+        Icrc21SupportedStandard {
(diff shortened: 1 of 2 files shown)


reference-implementations/ICRC-21/src/lib.rs
```diff
@@ -15,11 +15,17 @@ fn greet(name: String) -> String {
...
 #[ic_cdk::query]
-fn icrc21_supported_standards() -> Vec<Icrc21SupportedStandard> {
-    vec![Icrc21SupportedStandard {
-        url: "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md".to_string(),
-        name: "ICRC-21".to_string(),
-    }]
+fn icrc10_supported_standards() -> Vec<Icrc21SupportedStandard> {
+    vec![
+        Icrc21SupportedStandard {
+            url: "https://github.com/dfinity/ICRCs/ICRC-10.md".to_string(),
```
lmuntaner, line 21: This should be a valid one before we merge this, right?

frederikrothenberger: No, I wouldn't wait for that. Currently, this is an implementation of ICRC-10 as specified. The URL is anyway not as strictly specified (see [here](https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-10/ICRC-10.md#url)).

I think it is more important to adjust the method name now, as to not confuse people looking for guidance.

lmuntaner (approved):

reference-implementations/ICRC-21/src/lib.rs
lmuntaner: Ok, thanks for the explanation

PR #178, opened 2024-05-15
ICRC-29: Switch from signer notification to polling

frederikrothenberger:
By changing from the notification to polling, the signer no longer relies on the `window.opener` to be available to signal its readiness.

This is allows the signer window to redirect to other domains before signalling ready.

### DIFF
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -20,17 +20,30 @@ For this standard to represent an [ICRC-25](https://github.com/dfinity/wg-identi
 
 ## Establishing a Communication Channel
 
-A [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) communication channel is initiated by the relying party. The relying party opens a new window and waits for the signer to send a message indicating that it is ready for interactions.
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
LGTM as discussed in the working group

seahorsepip:
@frederikrothenberger 

> This is allows the signer window to redirect to other domains before signalling ready.

How would the relying party know that the incoming messages are coming from the correct window? Normally when a relying party opens a (popup) window for origin X you start listening for messages from origin X and send messages to origin X. 

But if that window redirects from origin X to origin Y, how would the relying party know that it should listen and send messages to origin Y instead of X?

Sending messages without specifying the origin leaks the message to all windows (including iframes) and listening to all messages without checking the origin is insecure since now any incoming window message is interpreted (including iframes).

sea-snake:
@frederikrothenberger

This is allows the signer window to redirect to other domains before signalling ready.

How would the relying party know that the incoming messages are coming from the correct window? Normally when a relying party opens a (popup) window for origin X you start listening for messages from origin X and send messages to origin X.

But if that window redirects from origin X to origin Y, how would the relying party know that it should listen and send messages to origin Y instead of X?

Sending messages without specifying the origin leaks the message to different origins (in case the window has navigated to a different origin) and listening to all messages without checking the origin is insecure since now any incoming window message is interpreted (including iframes).

PR #184, opened 2024-06-13
ICRC-21: Add UTC offset to consent message preferences

frederikrothenberger:
In the ongoing implementation of consent messages in the ledger canisters timezone conversion was raised as an issue.

An optional UTC offset is added to the consent message metadata to address that requirement.


sea-snake (approved):
Looks good to me.

PR #186, opened 2024-06-18
Rephrase ICRC-28 and -34

dostro:


topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+
+App-Specific Delegations **MUST** be restricted by the signer to individual relying parties.
+
+Signers **SHOULD** give users the choice to authenticate with their wallets (if available) or with an app-specific identifier.
+
+Relying parties **SHOULD** remove targets if they want to isolate users from the rest of the ecosystem.
+
+## Method
+
+**Name and Scope:** `icrc34_get_delegation`
```
frederikrothenberger, line 40:
```suggestion
**Name and Scope:** `icrc34_delegation`
```

frederikrothenberger (commented):
As per WG discussion, rename

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+    "maxTimeToLive": "28800000000000"
+  }
+}
+```
+
+### Example RPC Get Delegation Response (wallet delegation)
+
+`publicKey`: Public key of delegation identity as described in
+the [IC interface specification, signatures section](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).  
+`delegation`: An array of delegations (as defined by
```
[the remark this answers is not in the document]
sea-snake: @frederikrothenberger  @dostro  Small detail, looking at other specs, the key should probably be camelcase `signerDelegation`.

topics/icrc_28_trusted_origins.md
```diff
@@ -6,107 +6,63 @@
...
-
-```
-icrc28_supported_standards : () -> (vec record { name : text; url : text }) query;
-```
-
-The result should always have at least one entry:
-
-```
-record { name = "ICRC-28"; url = "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-28/ICRC-28.md" }
+icrc28_trusted_origins : () -> (record { trusted_origins : vec text });
```
frederikrothenberger, line 27: I feel like we will hit similar concerns as with alternative origins, that this will be mis-used for centralization. I think we should put a limit on how many entries can be returned here as well.

topics/icrc_28_trusted_origins.md
```diff
@@ -6,107 +6,63 @@
...
     * [Cold Signer Use-Case](#cold-signer-use-case)
 <!-- TOC -->
 
 ## Summary
 
-When a relying party wants to authenticate as a user, it uses
-a [session key](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures), and by way of the
-authentication flow ([ICRC-34](./icrc_34_get_global_delegation.md) or [ICRC-57](./icrc_57_get_session_delegation.md))
-obtains a delegation chain that allows the session key to sign for the user's global or session identity.
+This standard describes how a canister can indicate that a relying party is trusted.
```
frederikrothenberger, line 18: The term relying party is not defined in this context. Also, the semantics of trusted are really unclear.

The longer I think about this, the more I'm convinced that we should make the endpoint use-case specific (i.e. session isolation) rather than a generic "trusted" one.

topics/icrc_28_trusted_origins.md
```diff
@@ -6,107 +6,63 @@
...
-```
-record { name = "ICRC-28"; url = "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-28/ICRC-28.md" }
+icrc28_trusted_origins : () -> (record { trusted_origins : vec text });
 ```
 
 ## Use-Cases
 
-The `icrc28_get_trusted_origins` method is designed to be used with both cold and hot signers.
+There are several reasons why signers might want to get a list of frontend URLs a canister trusts, with one in 
+particular related to removing wallet approval prompts for a user's wallet address.
```
frederikrothenberger, line 33: I think the removal of wallet prompts is a second order effect. The main goal is to achieve session isolation _with_ a stable principal without signer interactions.

topics/icrc_28_trusted_origins.md
```diff
@@ -6,107 +6,63 @@
...
 
-### Hot Signer Use-Case
 
-This section describes the interactions between the signer and the relying party for _hot_ signers:
+### Wallet Delegation Use-Case
+
+[ICRC-34](./icrc_34_get_delegation.md) applies when a user returns a delegation to a relying party that 
+can be used to make authenticated calls on the user's behalf (i.e. without displaying wallet approval prompts).
```
frederikrothenberger, line 39:
```suggestion
can be used to make authenticated calls on the user's behalf (i.e. without interactions with the signer). 
```

topics/icrc_28_trusted_origins.md
frederikrothenberger: Also, this should come with some warning / sentence that the developer must ensure that all entries in the `trusted_origins` are not malicious and will not turn malicious in the future. I.e. by either having direct control of that origin, having the origin under DAO control, or ensuring that control over the origin is black-holed.
(Although, it is not recommended to black-hole a canister that serves front-end assets due to the unstable nature of the web environment).

Also, canisters that implement that interface should/must ensure that the list can be changed, in case an origin is no longer trusted.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+    * [Example RPC Get Delegation Request](#example-rpc-get-delegation-request)
+    * [Example RPC Get Delegation Response (wallet delegation)](#example-rpc-get-delegation-response-wallet-delegation)
+    * [Example RPC Get Delegation Response (app-specific delegation)](#example-rpc-get-delegation-response-app-specific-delegation)
+  * [Message Processing](#message-processing)
+  * [Errors](#errors)
+<!-- TOC -->
+
+## Summary
+
+When a relying party wants to remove wallet approval prompts and sign on behalf of its users, it uses the
```
frederikrothenberger, line 23: See comment above. I think we should talk here about removing interactions with the signer.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+for the relying party (App-Specific Delegation).
+
+Wallet Delegations **MUST** be restricted by the signer to canister targets that trust the relying party following the 
+[ICRC-28](./icrc_28_trusted_origins.md) standard. Doing otherwise would allow malicious actors to steal user assets.
+
+App-Specific Delegations **MUST** be restricted by the signer to individual relying parties.
+
+Signers **SHOULD** give users the choice to authenticate with their wallets (if available) or with an app-specific identifier.
+
+Relying parties **SHOULD** remove targets if they want to isolate users from the rest of the ecosystem.
```
frederikrothenberger, line 36:
```suggestion
Relying parties **May** remove the targets restriction if they issue an app-specific delegation.
```

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+When a relying party wants to remove wallet approval prompts and sign on behalf of its users, it uses the
+below `icrc34_get_delegation` method to obtain a delegation chain.
+
+The delegation chain's identifier may the user's wallet address (Wallet Delegation) or an address meant to be exclusively used 
+for the relying party (App-Specific Delegation).
+
+Wallet Delegations **MUST** be restricted by the signer to canister targets that trust the relying party following the 
+[ICRC-28](./icrc_28_trusted_origins.md) standard. Doing otherwise would allow malicious actors to steal user assets.
+
+App-Specific Delegations **MUST** be restricted by the signer to individual relying parties.
```
frederikrothenberger, line 32: As discussed in the working group, we should reconsider calling it app specific. Maybe rp-specific to keep the terminology consistent?

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Prerequisite:** ICRC-25 granted permission scope `icrc34_get_delegation` or `*`.
+
+## Scope (according to the [ICRC-25 standard](./icrc_25_signer_interaction_standard.md))
+
+**Scope:** `icrc34_get_delegation`
+
+**Required Properties:**
+
+- `publicKey` (`text` string): An Ed25519 or ECDSA session key as described [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).
```
frederikrothenberger, line 50:
```suggestion
- `publicKey` (`text` string): A DER encoded public key of a supported signature scheme, as described [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).
```

topics/icrc_34_get_delegation.md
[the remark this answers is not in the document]
frederikrothenberger: Agreed, thanks for spotting that.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+## Message Processing
+
+1. The relying party sends a `icrc34_get_delegation` request to the signer.
+2. Upon receiving the request, the signer validates whether it can process the message.
+    * If the relying party has not been granted the permission to request the action, the signer sends a response with
+      an error back to the relying party.
+    * The relying party must make sure that the request complies with scope targets restriction.
+3. If the request includes targets, the signer tries to retrieve and verify the trusted origins according to the [ICRC-28](icrc_28_trusted_origins.md) specification.
```
frederikrothenberger, line 183: The signer should be allowed to skip this step, if it does not offer wallet delegations at all.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Name and Scope:** `icrc34_get_delegation`
+
+**Prerequisite:** ICRC-25 granted permission scope `icrc34_get_delegation` or `*`.
+
+## Scope (according to the [ICRC-25 standard](./icrc_25_signer_interaction_standard.md))
+
+**Scope:** `icrc34_get_delegation`
+
+**Required Properties:**
```
frederikrothenberger, line 48: This is the wrong section: We are specifying the scope here and not the actual delegation request.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+    "scopes": [
+      {
+        "method": "icrc34_get_delegation"
+      }
+    ]
+  }
+}
+```
+
+## Requests
```
frederikrothenberger, line 76: I would structure this the same as e.g. ICRC-49: Have the schema / properties defined in this section, and then the example doesn't need to repeat that but can just be the JSON RPC request.

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+      "xhy27-fqaaa-aaaao-a2hlq-cai"
+    ],
+    "maxTimeToLive": "28800000000000"
+  }
+}
+```
+
+### Example RPC Get Delegation Response (wallet delegation)
+
+`publicKey`: Public key of delegation identity as described in
```
frederikrothenberger, line 102: This should be defined in a proper `Response` section rather than inline with the examples (which now is also duplicated).

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+## Requests
+
+### Example RPC Get Delegation Request
+
+Requests for a delegation can optionally include `targets`, enabling signers to verify the security of returning a wallet 
+delegation (via ICRC-28). Signers in this case **SHOULD** offer users the option to connect to the application with either 
+a wallet delegation (limited to canisters listed in `targets`) or an app-specific delegation (without targets). If the array
```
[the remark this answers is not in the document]
dostro: Will try "Account Delegation" and "Relying Party Delegation"

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+**Required Properties:**
+
+- `publicKey` (`text` string): An Ed25519 or ECDSA session key as described [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).
+
+**Optional Properties:**
+
+- `targets` (`text` array): A list of target canister ids (textual representation) the scope is restricted to.
+- `maxTimeToLive` (`text` array): Expiration of the delegation in nanoseconds, though signer can still choose to
+return a delegation with a shorter expiration.
+- `idleOptions`
```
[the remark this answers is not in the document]
dostro: Removed

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+## Message Processing
+
+1. The relying party sends a `icrc34_get_delegation` request to the signer.
+2. Upon receiving the request, the signer validates whether it can process the message.
+    * If the relying party has not been granted the permission to request the action, the signer sends a response with
+      an error back to the relying party.
+    * The relying party must make sure that the request complies with scope targets restriction.
+3. If the request includes targets, the signer tries to retrieve and verify the trusted origins according to the [ICRC-28](icrc_28_trusted_origins.md) specification.
+    * If the trusted origins cannot be retrieved for any of the given delegations targets or the relying party origin is
+      not within any of the trusted origin lists, the signer does not give users the ability to continue with the wallet delegation.
```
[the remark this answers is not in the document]
dostro: I would keep all-or-nothing because:
1. The only reason it wouldn't pass is if the app is being malicious or there's a mistake that should be corrected
2. There's more tooling required to keep track of which targets were not included in the returned delegation

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Scope:** `icrc34_get_delegation`
+
+**Required Properties:**
+
+- `publicKey` (`text` string): An Ed25519 or ECDSA session key as described [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).
+
+**Optional Properties:**
+
+- `targets` (`text` array): A list of target canister ids (textual representation) the scope is restricted to.
```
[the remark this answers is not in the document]
dostro: What specifically do you mean by if you "get"? Did you mean "if I request" or did you mean "if I receive back from the signer"?
I've updated to: "A list of target canister ids (textual representation) Account Delegations can execute transactions
for on behalf of the user."

topics/icrc_34_get_delegation.md
```diff
@@ -0,0 +1,192 @@
...
+
+1. The relying party sends a `icrc34_get_delegation` request to the signer.
+2. Upon receiving the request, the signer validates whether it can process the message.
+    * If the relying party has not been granted the permission to request the action, the signer sends a response with
+      an error back to the relying party.
+    * The relying party must make sure that the request complies with scope targets restriction.
+3. If the request includes targets, the signer tries to retrieve and verify the trusted origins according to the [ICRC-28](icrc_28_trusted_origins.md) specification.
+    * If the trusted origins cannot be retrieved for any of the given delegations targets or the relying party origin is
+      not within any of the trusted origin lists, the signer does not give users the ability to continue with the wallet delegation.
+4. The signer displays all the available delegations the user can continue with, and user selects one.
```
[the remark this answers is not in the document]
dostro: I think adding **MAY** in a couple locations clears it up well:
If the request includes targets, the signer **MAY** try to retrieve and verify the trusted origins according to the [ICRC-28](icrc_28_trusted_origins.md) specification.
The signer **MAY** display all the available delegations the user can continue with, in which case a user would select one.

topics/icrc_28_trusted_origins.md
dostro: Re: entry limit. This seems overly prescriptive IMHO and there's a performance limitation anyway as the array size grows.

Re: ensuring trusted_origins aren't malicious: Wouldn't these be outside the scope of a standard and more within scope of implementation? How about having this in the summary:

This standard describes how a canister can indicate that a relying party is trusted.

This standard does not describe how a canister:
- Should ensure that all entries are not malicious
- Should handle entries that become malicious

topics/icrc_28_trusted_origins.md
dostro: I'm not sure we want to be prescriptive on use cases. Why not let the community explore what other use cases they could build with this?

Here's a proposed edit:
This standard describes how a canister can indicate that a relying party (an entity that relies on the canister for certain functions or services) is trusted.

A trusted relying party carries certain privileges, like for example the ability to request Account Delegations as per ICRC-34.

This standard does not describe how a canister should:
- Ensure that all entries are free from malicious intent.
- Handle entries that become malicious over time.

topics/icrc_28_trusted_origins.md
dostro: I feel like these are the same goals, where one is more clear in the context of web3 than the other. No?

topics/icrc_34_get_delegation.md
dostro: Added "may"

topics/icrc_34_get_delegation.md
dostro: updated

topics/icrc_34_get_delegation.md
dostro: Relying parties aren't the ones issuing delegations. Did you mean to apply this to the sentence on signers?

topics/icrc_34_get_delegation.md
```diff
@@ -54,34 +57,38 @@ trust the relying party following the [ICRC-28](./icrc_28_trusted_origins.md) st
...
-the [IC interface specification, signatures section](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).  
-**`targets` (`text` array):** A list of target canister ids (textual representation) the scope is restricted to.  
-**`maxTimeToLive` (`text` array):** (Optional) Expiration of the delegation in nanoseconds, signer can still choose to
+An ICRC-25 compliant signer must implement the [icrc25_supported_standards](icrc_25_signer_interaction_standard.md#icrc25_supported_standards) method which returns the list of supported standards. 
+Any signer implementing ICRC-34 must include a record with the name field equal to "ICRC-34" in that list.
+
+## Request params
+
+- `publicKey` (`text`): A DER encoded public key of a supported signature scheme, as described [here](https://internetcomputer.org/docs/current/references/ic-interface-spec/#signatures).
+- `targets` (`text` `array` optional): A list of target canister ids (textual representation) Account Delegations can execute transactions
```
lmuntaner, line 75: If target canisters are mandatory, why is this parameter optional? What's the default?

> Account Delegations **MUST** be restricted by the signer to canister targets that trust the relying party

topics/icrc_28_trusted_origins.md
frederikrothenberger: I think, we want to be very specific with the concepts we have defined in the standards. The higher-level goal of not prompting the user IMHO is better placed in developer documentation, whereas here we should be very specific to what is actually happening.

topics/icrc_34_get_delegation.md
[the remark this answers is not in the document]
dostro: From WG meeting: signer MUST verify...

PR #200, opened 2024-07-16
docs: ICRC-29 formatting

peterpeterparker:
# Motivation

I had a bit of trouble understanding the "Signer" and "Relying party" chapters of ICRC-29. The content was clear, but it took a few reads to piece everything together.

So, here's a PR with minor formatting suggestions for those two chapters.

The PR also adds a reference to MDN for the `message` event.

# Changes

- Invert order of the chapters (because the communication is initiated by the relying party)
- Mark some comment as markdown comments
- Add a link to message on mdn
- Join two sentences


frederikrothenberger (approved):
Thanks, that makes sense.

PR #201, opened 2024-07-16
add derivationOrigin to 27 and 34

dostro:


sea-snake:
As mentioned on Discord, the question remains if this should be handled on:

- Transport layer and ICRC-34
- Transport layer only
- ICRC-34 only

I'm initially leaning towards transport layer only, a relying party can connect with origin x.com but identify as y.com through a yet unspecified transport layer spec. 

But there might be reasons why you'd want a more complicated approach of identifying as x.com but getting delegations for y.com.

I'm currently limited available for discussions but will be back in ~ a week.

frederikrothenberger:
> As mentioned on Discord, the question remains if this should be handled on:
> 
>     * Transport layer and ICRC-34
> 
>     * Transport layer only
> 
>     * ICRC-34 only
> 
> 
> I'm initially leaning towards transport layer only, a relying party can connect with origin x.com but identify as y.com through a yet unspecified transport layer spec.
> 
> But there might be reasons why you'd want a more complicated approach of identifying as x.com but getting delegations for y.com.
> 
> I'm currently limited available for discussions but will be back in ~ a week.

Ah, that's actually an interesting take on the issue. I didn't consider moving that to the transport layer so far...

I'll be on PTO the next 2 weeks, but we should definitely have a discussion about this in the next WG session on August 6.

PR #202, opened 2024-08-12
docs: typo granted in permissions

peterpeterparker:
# Motivation

There was two small typos in `icrc25_permissions`. It mentionned "grandted" permissions only while the response and standard actually list all permissions, not only those that are granted.


frederikrothenberger (approved):
Thanks a lot, LGTM!

PR #215, opened 2024-10-25
[Draft] [ICRC X] Content type of consent message

baolongt:
- This specification inherited functionalities from ICRC_21 standard 
- This specification return the consent message based on request and support by the canister

### DIFF
--- a/topics/ICRC-X/ICRC-X.did
+++ b/topics/ICRC-X/ICRC-X.did
@@ -1,0 +1,142 @@
+
+type icrc21_consent_message_metadata = record {
+    // BCP-47 language tag. See https://www.rfc-editor.org/rfc/bcp/bcp47.txt
+    language: text;
+
+    // The users local timezone offset in minutes from UTC.
+    // Applicable when converting timestamps to human-readable format.
+    //
+    // If absent in the request, the canister should fallback to the UTC timezone when creating the consent message.
+    // If absent in the response, the canister is indicating that the consent message is not timezone sensitive.
+    utc_offset_minutes: opt int16;
+};
+
+type icrc21_consent_message_spec = record {
+    // BCP-47 language tag. See https://www.rfc-editor.org/rfc/bcp/bcp47.txt
+    language: text;
+
+    // The users local timezone offset in minutes from UTC.
+    // Applicable when converting timestamps to human-readable format.
+    //
+    // If absent in the request, the canister should fallback to the UTC timezone when creating the consent message.
+    // If absent in the response, the canister is indicating that the consent message is not timezone sensitive.
+    utc_offset_minutes: opt int16;
+};
+
+// This inherit from ICRC 21 
+type icrcX_consent_message_request = record {
+    // Method name of the canister call.
+    method: text;
+    // Argument of the canister call.
+    arg: blob;
+    // User preferences with regards to the consent message presented to the end-user.
+    user_preferences: icrc21_consent_message_spec;
+    // Content type of the consent message.
+    content_type: text;
+};
+
+type icrcX_consent_message = variant {
+    // Message for a generic display able to handle large documents and do proper line wrapping and pagination / scrolling.
+    // Uses Markdown formatting, no external resources (e.g. images) are allowed.
+    GenericDisplayMessage: text;
+};
+
+
+type icrc21_consent_info = record {
+    // Consent message describing in a human-readable format what the call will do.
+    //
+    // The message should adhere as close as possible to the user_preferences specified in the consent_message_spec + format in content_type
+    // of the icrcX_consent_message_request.
+    // If the message is not available for the given user_preferences any fallback message should be used. Providing a
+    // message should be preferred over sending an icrc21_error.
+    // The metadata must match the consent_message provided.
+    //
+    // The message should be short and concise.
+    // It should only contain information that is:
+    // * relevant to the user
+    // * relevant given the canister call argument
+    // * formatted in valid content_type provided in the request
+    // Example with "content_type" : "application/json"
+    // {
+    //     "message": "This call will transfer 100 tokens from your account to the recipient account.",
+    //     "assetIn" : [
+    //                      {"address": "0x1234", "amount": 100}],     
+    //                  ]
+    //     "assetOut" : [
+    //                      {"address": "0x1234", "amount": 100}],
+    //                  ]   
+    //      "to": "0x1234",
+    //      "from": "0x1234",
+    //      "fee": 100,
+    // }
+    consent_message: icrcX_consent_message;
+    // Metadata of the consent_message.
+    metadata: icrc21_consent_message_metadata;
+
+};
+
+type icrc21_error_info = record {
+    // Human readable technical description of the error intended for developers, not the end-user.
+    description: text;
+};
+
+type icrcX_error = variant {
+    // The canister does not support this call (i.e. it will lead to a rejection or error response).
+    // Reasons might be (non-exhaustive list):
+    // * the canister call is malformed (e.g. wrong method name, argument cannot be decoded)
+    // * the arguments exceed certain bounds
+    //
+    // The developer should provide more information about the error using the description in icrc21_error_info.
+    UnsupportedCanisterCall: icrc21_error_info;
+
+    // The canister cannot produce a consent message for this call.
+    // Reasons might be (non-exhaustive list):
+    // * it is an internal call not intended for end-users
+    // * the canister developer has not yet implemented a consent message for this call
+    //
+    // The developer should provide more information about the error using the description in icrc21_error_info.
+    ConsentMessageUnavailable: icrc21_error_info;
+
+    // The canister did not provide a consent message for because payment was missing or insufficient.
+    //
+    // This error is used to account for payment extensions to be added in the future:
+    // While small consent messages are easy and cheap to provide, this might not generally be the case for all consent
+    // messages. To avoid future breaking changes, when introducing a payment flow, this error is already introduced
+    // even though there no standardized payment flow yet.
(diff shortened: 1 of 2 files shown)


sa-github-api:
Dear @baolongt,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/522) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

sea-snake:
Would be interested to hear the use cases of the content type added in this ICRC-21 extension. 

The main purpose of the ICRC-21 standard is to "Return a human-readable consent message for the given canister call e.g. transfer 10 ICP to XYZ". In the standard, this message can be returned as either rich markdown (browser wallet) or plain text for line displays (e.g. HW wallet).

How does content type relate here? I see in your MR an example with a JSON structure that contains both a plain text message and the transaction details. For the common use case of an (ICRC standardized) token transfer you wouldn't need ICRC-21 to get the transaction details, instead you could decode the candid arguments if the method name is recognized as e.g. icrc1_transfer.

Just to clarify, the ICRC-21 standard is primarily meant to generate consent messages for canister calls that are not standardized ICRC contract calls like e.g. icrc1_transfer. This enables for e.g. HW wallets to approve any canister call even if the call is not a standardized call like e.g. icrc1_transfer.

If you have time, also consider joining us in the bi-weekly Identity & Wallet Standard WG meetings. The agenda with zoom links can be found here: https://calendar.google.com/calendar/u/0/embed?src=c_cgoeq917rpeap7vse3is1hl310@group.calendar.google.com&ctz=Europe/Zurich

baolongt:
> Would be interested to hear the use cases of the content type added in this ICRC-21 extension.
> 
> The main purpose of the ICRC-21 standard is to "Return a human-readable consent message for the given canister call" e.g. transfer 10 ICP to XYZ. In the standard, this message can be returned as either rich markdown (browser wallet) or plain text for line displays (e.g. HW wallet).
> 
> How does content type relate here? I see in your MR an example with a JSON structure that contains both a plain text message and the transaction details. For the common use case of an (ICRC standardized) token transfer you wouldn't need ICRC-21 to get the transaction details, instead you could decode the candid arguments if the method name is recognized as e.g. icrc1_transfer.
> 
> Just to clarify, the ICRC-21 standard is primarily meant to generate consent messages for canister calls that are not standardized ICRC contract calls like e.g. icrc1_transfer. This enables for e.g. HW wallets to approve any canister call even if the call is not a standardized call like e.g. icrc1_transfer.
> 
> If you have time, also consider joining us in the bi-weekly Identity & Wallet Standards WG meetings. The agenda with zoom links can be found here: https://calendar.google.com/calendar/u/0/embed?src=c_cgoeq917rpeap7vse3is1hl310@group.calendar.google.com&ctz=Europe/Zurich

You're right. I just want to extend the ICRC-21 but didn't have any vision about what should schema it so I just put all the stuff in consentMassge

baolongt:
But in my case, I really need a similar things to consent message but showing what asset user need to send and what it is will return. I think it's best fit with JSON and thinking about content_type for ICRC_21

sea-snake:
>  showing what asset user need to send and what it is will return

Could you clarify this further, possibly with an example with how this would be used?

baolongt:
> > showing what asset user need to send and what it is will return
> 
> Could you clarify this further, possibly with an example with how this would be used?

You can see in the image
Screenshot 2024-10-25 at 21 59 01
Screenshot 2024-10-25 at 21 59 36

baolongt:
> > showing what asset user need to send and what it is will return
> 
> Could you clarify this further, possibly with an example with how this would be used?
Screenshot 2024-10-25 at 22 00 53

sea-snake:
## Markdown

Firstly wanted to clarify that text in combination with tables is possible with (extended) markdown. Keep in mind that the signer is responsible for rendering the markdown, so details like the table widths being equal and rounded corners might vary between signers.

### Design 1

```markdown
**You will receive**

|       |    Amount |
| ----- | --------: |
| Asset |     1 ICP |
| Asset | 15 ckUSDC |
| Asset |  100 BOOP |

**You will send**

|                       |     Amount |
| --------------------- | ---------: |
| Network fee ($0.0008) | 0.0001 ICP |
| Total fees ($0.0008)  | 0.0001 ICP |
```

**You will receive**

|       |    Amount |
| ----- | --------: |
| Asset |     1 ICP |
| Asset | 15 ckUSDC |
| Asset |  100 BOOP |

**You will send**

|                       |     Amount |
| --------------------- | ---------: |
| Network fee ($0.0008) | 0.0001 ICP |
| Total fees ($0.0008)  | 0.0001 ICP |

### Design 2

```markdown
| Swap     |               |
| -------- | ------------: |
| USD Coin | +1,00053 USDC |
| Solana   | -0,005272 SOL |

| Fees          |             |
| ------------- | ----------: |
| Mang luoi     |      Solana |
| Phi mang luoi | 0,00013 SOL |
```

| Swap     |               |
| -------- | ------------: |
| USD Coin | +1,00053 USDC |
| Solana   | -0,005272 SOL |

| Fees          |             |
| ------------- | ----------: |
| Mang luoi     |      Solana |
| Phi mang luoi | 0,00013 SOL |

## Data instead of message?

As far as I understand, this PR returns a JSON structure (data) instead of Markdown (message). The given wallet is expected to interpret the JSON structure and render an interface .

Besides the question whether this should be done by adding an asset type or another device type besides the `GenericDisplayMessage` (markdown) and `LineDisplayMessage` (plain text, paged).

The main question I have here, is how would a signer know how to render said JSON (data) to the screen? Keep in mind that the signer does not know what kind of call it's supposed to be rendering on the screen. The signer could receive for example a call to: transfer ICP, swap tokens, delete an account, sign up for an account, etc.

In the given two examples above there's already two different transaction types with different designs. As far as I can see from the design, the first one shows multiple assets being received while the second design is a swap. 

The main purpose of ICRC-21 is to receive a consent message (in human language) that the signer can render "as is" without prior knowledge of the type of canister call that's being made. 

In case of JSON (data), the signer would need to know all the various possible JSON structures and implement an interface for each one. Which extends on my earlier question, how would a signer maintain and keep adding support for this growing list of JSON structures?

## ICRC canister calls

In case of canisters calls that follow the ICRC standards, a signer could add a custom implementation for a given standard e.g. ICRC-1. 

This custom implementation would recognize that a call is an ICRC-1 transfer and as a result decode the candid arguments and use these arguments to render a custom interface directly.

An example of decoding ICRC-1 transfer arguments can be found [here](https://github.com/slide-computer/tokens-js/blob/main/packages/tokens-icrc1/src/index.ts#L88).

The purpose of ICRC-21 is to primarily handle all the canister calls that are not standardized, and thus the signer does not have a decode and custom interface implementation for.

Standardizing swaps and other related common transactions in ICRC standard(s) would enable signers to add support for decoding their arguments and rendering a custom interface.

# TL;DR

- How would a signer know how to render said JSON (data) to the screen?
- How would a signer maintain and keep adding support for this growing list of JSON structures?

baolongt:
1. How would a signer know how to render said JSON (data) to the screen?

The ICRC21 is programable for consent message. So the JSON returns will be a public schema for all develop integrate with the canister. I think this will lead to another topic that how should the schema is.

2. How would a signer maintain and keep adding support for this growing list of JSON structures?

Do you think we should define a JSON for general purpose? Can we define a ICRC for a JSON schema?

sea-snake:
> So the JSON returns will be a public schema for all develop integrate with the canister.

Isn't that what the candid definition of a canister is meant for?

The candid definition of a canister is basically a public schema of the canister and all of it's calls and arguments.

> Can we define a ICRC for a JSON schema?

Why not define a ICRC standard for the candid definition of said canisters(s) like ICRC-1 and other ICRC standards do?
 
The added benefit of this existing standardization process is that it also documents the canister calls themselves for other canisters besides signers.

Just to clarify the purpose of ICRC-21 is to return a consent message, something a signer can always display without any prior knowledge of any other ICRC standards. 

If ICRC-21 returns something other than a consent message, for example a JSON schema. Then I'm not sure if this is still within the scope of ICRC-21, since it's no longer a generic message that can be rendered without prior knowledge.

Also in case of a swap platform, it probably won't need ICRC-21 since it could use ICRC-34 delegations to avoid the need of going through the signer when calling its own platform canister(s). So above swap examples will be able to show a well designed confirmation popup within the dapp itself without any signer approval screens needed.

baolongt:
So this mean my ideal (picture 1) is like just in my dapp problem to show consent message before claim something rather than a general solution for canister communication from signer?

sea-snake:
> So this mean my ideal (picture 1) is like just in my dapp problem to show consent message before claim something rather than a general solution for canister communication from signer?

In practice a swap flow could look something like this:

1. Request signer to make icrc2_approve call for token Y held by signer. Signer shows user that the dapp wants to control X funds for token Y on behalf of the user.
2. User configures swap in dapp
3. Dapp shows swap details and user confirms.

Only in step 1 the user has to approve in the signer, step 2 and 3 are all within the dapp itself. 

Ideally you'd only have signer interaction to access the user funds (ledger canisters), for interaction with the dapp itself (dapp own canisters), there's no need for signer interaction.

The purpose of ICRC-21 is to have something (a message) to show in case there's no other choice besides signer interaction and the canister call is not ICRC standardized like e.g. an ICRC-1 transfer of ICRC-2 approval. Ideally you'd try to avoid going through the signer for interaction with the dapp itself.

PR #216, opened 2024-10-27
Rephrase and add heartbeats to the ICRC-29 standard.

sea-snake:
As mentioned in earlier WG calls, relying on the window close status would be unreliable. The solution was proposed to use heartbeats to indicate to both the relying party and signer that they're available and ready to receive messages.

This PR updates the ICRC-29 standard to implement heartbeats with minimal changes from the signer side and removes the no longer relevant session references to ICRC-25.

### DIFF
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -30,7 +30,7 @@ The message has `targetOrigin` set to `'*'` and is a [JSON-RPC 2.0 (https://www.
 }
 ```
 
-Expected response from signer:
+The signer should send responses with the `targetOrigin` set to the previously received `icrc29_status` message its `origin` property value:
 ```json
 {
     "jsonrpc": "2.0",
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -24,18 +24,18 @@ The message has `targetOrigin` set to `'*'` and is a [JSON-RPC 2.0 (https://www.
 
 ```json
 {
-  "jsonrpc": "2.0",
-  "id": "1",
-  "method": "icrc29_status"
+    "jsonrpc": "2.0",
+    "id": "1",
+    "method": "icrc29_status"
 }
 ```
 
 Expected response from signer:
 ```json
 {
-  "jsonrpc": "2.0",
-  "id": "1",
-  "result": "ready"
+    "jsonrpc": "2.0",
+    "id": "1",
+    "result": "ready"
 }
 ```
 
--- a/topics/icrc_29_window_post_message_transport.md
+++ b/topics/icrc_29_window_post_message_transport.md
@@ -18,7 +18,7 @@ For this standard to represent an [ICRC-25](https://github.com/dfinity/wg-identi
 
 ## Communication Channel
 
-The relying party initiates and maintains the communication channel by opening a new window for the signer and periodically sending `icrc29_status` messages using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) API.
+The relying party initiates and maintains the communication channel by opening a new window for the signer and periodically sending `icrc29_status` messages using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) API to indicate it is ready for interactions.
 
 The message has `targetOrigin` set to `'*'` and is a [JSON-RPC 2.0 (https://www.jsonrpc.org/specification) call with method `icrc29_status` and no parameters:
 
@@ -30,7 +30,7 @@ The message has `targetOrigin` set to `'*'` and is a [JSON-RPC 2.0 (https://www.
 }
 ```
 
-The signer should send responses with the `targetOrigin` set to the previously received `icrc29_status` message its `origin` property value:
+The signer should respond to every `icrc29_status` messsage received to indicate it is ready to receive additional messages. The response message `targetOrigin` should be set to the received `icrc29_status` message its `origin` property value:
 ```json
 {
     "jsonrpc": "2.0",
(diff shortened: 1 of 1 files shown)


topics/icrc_29_window_post_message_transport.md
```diff
@@ -16,10 +16,11 @@ For this standard to represent an [ICRC-25](https://github.com/dfinity/wg-identi
 * The user's machine is not compromised. In particular, the browser must deliver messages unchanged and represent the `origin` of the different parties correctly.
 * DNS entries are not compromised. The user's machine must be able to resolve the domain names of the relying party and signer correctly.
 
-## Establishing a Communication Channel
+## Communication Channel
 
-A [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) communication channel is initiated by the relying party. The relying party opens a new window and polls repeatedly for the signer to reply with a message indicating that it is ready for interactions.
-The message sent by the relying party has `targetOrigin` set to `'*'` and is a [JSON-RPC 2.0](https://www.jsonrpc.org/specification) call with method `icrc29_status` and no parameters:
+The relying party initiates and maintains the communication channel by opening a new window for the signer and periodically sending `icrc29_status` messages using the [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) API.
```
peterpeterparker, line 21: We loose the notion that the signer is supposed to reply with a message indicating that it is ready for interactions. Is this expected?

topics/icrc_29_window_post_message_transport.md
```diff
@@ -29,9 +30,7 @@ The message sent by the relying party has `targetOrigin` set to `'*'` and is a [
 }
 ```
 
-The signer should reply with a message indicating that it is ready to receive additional messages.
-The message `targetOrigin` should be set to the previously received `icrc29_status` message its `origin` property value.
-
+The signer should send responses with the `targetOrigin` set to the previously received `icrc29_status` message its `origin` property value:
```
peterpeterparker, line 33: ditto, no more mention of ready?

I would also maybe use singular - "The signer should respond" - instead of "responses". I was confuse if it meant that the signer should respond with multiple responses. WDYT?

topics/icrc_29_window_post_message_transport.md
```diff
@@ -40,10 +39,16 @@ The message `targetOrigin` should be set to the previously received `icrc29_stat
 }
 ```
 
-> **Note**: The relying party should send `icrc29_status` messages in short intervals. It is expected that some of the messages will be lost due to being sent before the signer is ready.
+### Establishment
+
+The connection is considered established once the relying party receives a `"result": "ready"` response to an `icrc29_status` message, which may not necessarily be the first message sent. In case the relying party does receive any `"result": "ready"` response within a reasonable timeframe, it should treat this as a failure to establish the connection.
```
peterpeterparker, line 44: - "which may not necessarily be the first message sent.": I'm not sure what that mean. Do you mean sent by the relying party or the signer?

- "In case the relying party does receive": isn't the keyword NOT missing here?

topics/icrc_29_window_post_message_transport.md
```diff
@@ -40,10 +39,16 @@ The message `targetOrigin` should be set to the previously received `icrc29_stat
...
+
+Once the connection is established, the relying party should continue sending `icrc29_status` messages at regular intervals to maintain the connection.
```
peterpeterparker, line 46: "could" instead of "should"?

Polling to maintain the connection open is not a must or, is it?

topics/icrc_29_window_post_message_transport.md
```diff
@@ -70,15 +75,14 @@ Messages are received by listening to `message` events and are considered as com
...
-The window must not be automatically closed after sending a message, since the relying party could possibly send additional messages. 
-Instead, the relying party is responsible for closing the signer window.
+The window must not be automatically closed after sending a message, since the relying party could 
+possibly send additional messages over the existing communcation channel. Instead, the relying party is responsible for closing the signer window.
 
 ## Error Handling
 
 ### Unexpectedly Closed Window
 
-If the signer window is closed unexpectedly, the relying party should handle this as if it had received a `NOT_GRANTED` error.
```
peterpeterparker, line 80: Any particular reason to have removed this reference to `NOT_GRANTED` error?

topics/icrc_29_window_post_message_transport.md
sea-snake: It's a must, else a dapp would need to implement multiple approaches to keep track of a connection still being open instead of the single approach described here.

I changed it from should to must to be more clear in this regard.

topics/icrc_29_window_post_message_transport.md
sea-snake: The JSON RPC messages (either a response or an error) are meant as response to other JSON RPC messages. The communication channel being closed is not an error in response to a JSON RPC message and thus this line has been removed to resolve that.

Instead the communication channel being closed is a ICRC-29 specific transport channel error that should be handled outside the scope of the data sent through this transport channel.

topics/icrc_29_window_post_message_transport.md
sea-snake: You're right, added it back

topics/icrc_29_window_post_message_transport.md
sea-snake: Clarified why this may not be a response to the first message sent.

And yes, "not" was missing there.

topics/icrc_29_window_post_message_transport.md
sea-snake: You're right, added it back.

Clarified that the signer should respond to every message.

peterpeterparker (approved):
LGTM, thx

PR #220, opened 2024-11-25
[Draft] [ICRC X] Batch canisters call

baolongt:


topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+- [ICRC-X: Call batch canisters](#icrc-x-call-batch-canisters)
+  - [Summary](#summary)
+  - [Processing](#processing)
+    - [Examples](#examples)
+      - [Successful Batch Call](#successful-batch-call)
+      - [Batch Call with Error](#batch-call-with-error)
+
+## Summary
+
+This Method can be used by the relying party to request batch call to 3rd party canister executed by the signer using the requested identity. In order to prevent misuse of this method all `icrcX_call_batch_canisters` requests are subject to user approval.
```
sea-snake, line 18:
```suggestion
This method can be used by the relying party to request a batch call to 3rd party canister executed by the signer using the requested identity. In order to prevent misuse of this method all `icrcX_call_batch_canisters` requests are subject to user approval.
```

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+
+There is two main parameter for this standard
+
+1. `mode` : The mode to execute transaction in sequence or in parallel
```
sea-snake, line 22: Instead of picking between a mode, wouldn't it be more flexible to combine modes?

For example a user might want to ask 2 approvals in parallel before making a call to a swap. To do something like that I'd either nest requests inside requests of for even more flexibility, add a dependency array to each request that indicates which requests must be finished before the request is executed.

An example of optional dependency arrays would look something like this.

```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "icrcX_batch_call_canisters",
  "params": {
    "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
    "requests": [
      {
        "id": 1,
        "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
        "method": "approve",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
      },
      {
        "id": 2,
        "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
        "method": "approve",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
      },
      {
        "id": 3,
        "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
        "method": "swap",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
        "waitFor": [1, 2],
      },
      {
        "id": 4,
        "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
        "method": "swap",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
        "waitFor": [3],
      }
    ]
  }
}
```

Requests would execute in this example: `1 & 2 -> 3 -> 4`

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+2. `requests` : An array of canister call requests
+
+Unfortunately there is currently no mode to execute all transactions in the batch in one block. The transactions are handled as independent transactions that are executed in parallel or in sequence. However, ICRCX allows signer to receive approval for all transactions in the batch from the user with one approval action.
```
sea-snake, line 25: Since it's a single approval action and we want to be sure that the user is fully aware and have an overview of what it's approving (e.g. read all consent messages). Should there be a maximum limit of nr of calls?

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+1. Relying party requests a ICRCX batch transaction request to the signer. The request can be either `parallel` mode or `sequence` mode (see more details below). 
+
+2. The signer fetches consent messages and shows a warning to user to approve the batch transaction. 
+
+3. The signer submits canister calls to target canisters:
+- mode `parallel`: Execute all requests simultaneously, without waiting for any individual request to complete before starting the next. All transactions will be started. But it is not guaranteed that all transactions will succeed.
+- mode `sequence`: 
+  - Execute each request one after the other, ensuring that each request is completed before starting the next one. If any transaction fails, the execution of the batch will stop without executing remaining queued transactions. 
+  - Evaluate the result of processing the request:
+     1. If the result is a successful response, add the result to the batch response. Proceed with processing the next request in the batch.
```
sea-snake, line 38: As mentioned in the WG call and also Discord, we can't see if the call itself is successful regarding canister logic e.g. success or error result. But we can see (in the certificate) if the call has successfully [received a reply](https://github.com/dfinity/agent-js/blob/main/packages/agent/src/polling/index.ts#L65) from the canister or if something went wrong in between. 

This reply can be any byte array, in practice this is always cbor that will be decoded by the agent-js actor with knowledge of the candid definition of the canister. But even decoded from cbor into js values, there is no clear indicator of success or error, the value that indicates success or error is up to every canisters implementation itself to decide. It could be a boolean, enum or number range that indicates success or error.

So all we can do within this limitation is to stop the chain of calls when:
- No reply has been received
- Something else went wrong in between e.g. IC could not be reached

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "responses": [
+      {
+        "id": 1,
+        "result": {
+          "Ok": "16509194"
```
sea-snake, line 133:
```suggestion
          "contentMap": "2dn3p2NhcmdYTkRJREwEbXtuAGwCs7DawwNorYbKgwUBbAP7ygECot6U6wYB2KOMqA19AQMBHVfs7SoKuxOJdX37k7sBJsPALWEo5ayflizuaH0CAADIAWtjYW5pc3Rlcl9pZEoAAAAAAcDR1wEBbmluZ3Jlc3NfZXhwaXJ5GxeNX/65y4YAa21ldGhvZF9uYW1laHRyYW5zZmVyZW5vbmNlUFF4+hAimFhoqkdUcIchz0xscmVxdWVzdF90eXBlZGNhbGxmc2VuZGVyWB1q63Snu+4C5/fpWFu4nq1IpZxCYDEYA8XSPqPfAg==",
          "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
```

As shown [here](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/icrc_49_call_canister.md#example-rpc-response) in ICRC-49, the signer can't and shouldn't decode the response here since it doesn't know the response encoding and the relying party needs the original certificate and contentMap to verify the call has actually been made as requested.

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+```
+
+## Errors
+
+In addition to the errors defined in [ICRC-25](./icrc_25_signer_interaction_standard.md#errors-3) this standard defines the following errors:
+
+
+| Code  | Message                                    | Meaning                                                                                          | Data                                                                                |
+|-------|--------------------------------------------|--------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|
+| 10101 | Not processed due to batch request failure | The message was not processed as one of the preceding request in the batch resulted in an error. | (optional) Error details: <ul> <li>`message` (`text`, optional): message</li> </ul> |
```
sea-snake, line 237: Error codes have been [updated in ICRC-49](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/icrc_25_signer_interaction_standard.md#errors-3) a while back, I would recommend to use error code 1001 since it doesn't fit within 2xxx, 3xxx or 4xxx categories.

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,237 @@
...
+                "e8s": "0"
+              }
+            }
+          }
+        }
+      },
+      {
+        "id": 3,
+        "error": {
+          "code": 1000,
```
sea-snake, line 221: Would be nice to also have an example where the processing of calls has stopped in between after an error with a call:

- result -> ...
- result -> ...
- error -> code: 4000
- error -> code: 1001
- error -> code: 1001

topics/icrc_X_multiple_canister_call.md
baolongt: but how do we know the dependencies 1 or 2 are success or not in order to execute 3

baolongt: So from the question in discord 

```
If we really want to implement sequence, we could ask the relying party to pass the expected success response, such as 'ok' (i..e  ICRC 1 or 2 ) or a non-null blockid, for each transaction as parameters to signer, and signer continues to execute the next tx only if the success code is received.
```

Can this possible to do? 
Based on your answer is yes for signer parsing data
But how about relying party passing `expected success response` to the signer like regex or maybe a full payload

sea-snake:
I was thinking that since the relying party provides the ICRC-21 inspired canister, the signer shouldn't trust it regarding communication with the user. 

For example this canister could return an error, but the error message shouldn't be shown to the user, since it could be false information. It should only be used to decide if the sequence should continue or not.

But while thinking this, it could actually be interesting to consider the possibility of the target canister (additionally opt in) providing the same success/error protocol itself. In that case, the signer can trust and display the success/error message. 

This would probably be another standard, but it could also be the same standard, just implemented by either the relying party (untrusted, don't show details to user) or the target canister (trusted, show details to user).

In the trusted example, a signer could show a consent message "would you like to transfer X tokens" and afterwards also show a message "transfer succeeded" or "transfer failed".  This would be an additional security improvement over trusting the relying party to tell you what happened.

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,331 @@
...
+        "waitFor": ["1", "2"]
+      },
+      ,
+      {
+        "id": "4",
+        "canisterId": "bbbbb-fqaaa-aaaao-a2hlq-ca",
+        "method": "bridge_to_eth",
+        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB",
+        "waitFor": ["3"],
+        "validateCanister": "xyzzz-fqaaa-aaaao-a2hlq-ca"
```
sea-snake, line 186: Assuming that the validateCanister transform the response from request 4 into success true/false, it doesn't have any affect when defined in the last request since either way no other calls are executed after the last call.

So it should probably be moved to request 3 (maybe also 1 and 2?).

Is there a use case for having a different validateCanister for requests within the same batch? If not it might make more sense to move it up below the sender, this simplifies the signer implementation and enables it to validate multiple responses at once.

```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "icrcX_batch_call_canisters",
  "params": {
    "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
    "validate": {
      canisterId: "xyzzz-fqaaa-aaaao-a2hlq-cai"
      method: "validate_batch_approve_swap_bridge"
    },
    "requests": [
      {
        "id": "1",
        "canisterId": "eeddf-fqaaa-aaaao-a2hlq-ca",
        "method": "icrc2_approve",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
      },
      {
        "id": "2",
        "canisterId": "aaabb-fqaaa-aaaao-a2hlq-ca",
        "method": "icrc2_approve",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
      },
      {
        "id": "3",
        "canisterId": "xyzzz-fqaaa-aaaao-a2hlq-ca",
        "method": "swap",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB",
        "waitFor": ["1", "2"]
      },
      ,
      {
        "id": "4",
        "canisterId": "bbbbb-fqaaa-aaaao-a2hlq-ca",
        "method": "bridge_to_eth",
        "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB",
        "waitFor": ["3"]
      }
    ]
  }
}
```

The `validate_batch_approve_swap_bridge` method of canister `xyzzz-fqaaa-aaaao-a2hlq-cai` would be called twice, first with an array containing the responses from request 1 and 2 and later with an array containing the response from request 3. Keeping latency to a minimum by limiting the number of validation calls to the number of unique `waitFor` combinations.

baolongt: It's like ICRC21 for response registry canister. Yeah ok, I will talk this with Mason

baolongt: Is there any flag to not making a new call to validate the response?

If I'm a developer making the swap and bridge, I should do the validation for allowance before executing the swap and bridge. 

we can put `ignore` 
```
 "validate": {
      canisterId: "xyzzz-fqaaa-aaaao-a2hlq-cai"
      method: "validate_batch_approve_swap_bridge"
      ignore: ["3", "4"]
   },
```

or `validate` is optional here

sea-snake: I would argue that validating the last request is never needed, since there is no request after the last request that depends on its outcome.

In case you don't want to validate 3, then there's no reason to make 4 wait for 3 in the first place. You could argue that you'd want to make 4 wait for 3 independent of the outcome of 3 but that is an edge case I wouldn't cover here, instead simply always validate the responses other calls have waited for. Else we end up with more than two behaviors (parallel or sequential based on outcome) which makes things more complicated than needed.

As for making validate optional overall, I think that makes sense but then we'd need to decide what happens if there are waitFor properties defined, either:
- Throw error "validate is required when call depends on each other"
- Expect wallet to fallback to either defining success based on ICRC knowledge or "the call responded at least".

Considering that validation can be implemented without any reliance on devs that implemented the target canisters or wallets implementing ICRC knowledge correctly. I would argue to go for the first option, it guarantees consistent behavior across all signers since the dapp developer itself is in control.

sea-snake:
As discussed in WG meeting today:

- Remove ids from canister calls
- Group parallel canister calls as an array
- Return validation contentmap and certificate in validation error
- Add Validation required error in case there is more than one set of parallel calls (waiting for one another), since validation is required in this case.
- Also validate the last request (so all requests) when validation is defined

sea-snake:
Example JSON:
```json
{
  "request": {
    "id": 1,
    "jsonrpc": "2.0",
    "method": "icrcX_batch_call_canisters",
    "params": {
      "mode": "",
      "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
      "validation": {
         "canisterId": "xxxxx-xxxxx-xxxxx-xxxxx-xxx",
         "method": "validate_approve_swap_bridge"
      },
      "requests": [
        [
          {
            "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
            "method": "transfer",
            "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
          },
          {
            "canisterId": "xhy27-fqaaa-aaaao-a2hlq-ca",
            "method": "transfer",
            "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
          }
        ],
        {
          "canisterId": "xxxxx-fqaaa-aaaao-a2hlq-ca",
          "method": "swap",
          "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
        },
        {
          "canisterId": "yyyyy-fqaaa-aaaao-a2hlq-ca",
          "method": "bridge_to_eth",
          "arg": "RElETARte24AbAKzsNrDA2ithsqDBQFsA/vKAQKi3pTrBgHYo4yoDX0BAwEdV+ztKgq7E4l1ffuTuwEmw8AtYSjlrJ+WLO5ofQIAAMgB"
        }
      ]
    }
  },
  "response": {
    "id": 1,
    "jsonrpc": "2.0",
    "result": {
      "responses": [
        [
          {
            "result": {
              "contentMap": "2dn3p2NhcmdYTkRJREwEbXtuAGwCs7DawwNorYbKgwUBbAP7ygECot6U6wYB2KOMqA19AQMBHVfs7SoKuxOJdX37k7sBJsPALWEo5ayflizuaH0CAADIAWtjYW5pc3Rlcl9pZEoAAAAAAcDR1wEBbmluZ3Jlc3NfZXhwaXJ5GxeNX/65y4YAa21ldGhvZF9uYW1laHRyYW5zZmVyZW5vbmNlUFF4+hAimFhoqkdUcIchz0xscmVxdWVzdF90eXBlZGNhbGxmc2VuZGVyWB1q63Snu+4C5/fpWFu4nq1IpZxCYDEYA8XSPqPfAg==",
              "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
            }
          },
          {
            "result": {
              "contentMap": "2dn3p2NhcmdYTkRJREwEbXtuAGwCs7DawwNorYbKgwUBbAP7ygECot6U6wYB2KOMqA19AQMBHVfs7SoKuxOJdX37k7sBJsPALWEo5ayflizuaH0CAADIAWtjYW5pc3Rlcl9pZEoAAAAAAcDR1wEBbmluZ3Jlc3NfZXhwaXJ5GxeNX/65y4YAa21ldGhvZF9uYW1laHRyYW5zZmVyZW5vbmNlUFF4+hAimFhoqkdUcIchz0xscmVxdWVzdF90eXBlZGNhbGxmc2VuZGVyWB1q63Snu+4C5/fpWFu4nq1IpZxCYDEYA8XSPqPfAg==",
              "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
            }
          }
        ],
        {
          "error": {
            "code": 1002,
            "message": "Validation failed",
            "contentMap": "2dn3p2NhcmdYTkRJREwEbXtuAGwCs7DawwNorYbKgwUBbAP7ygECot6U6wYB2KOMqA19AQMBHVfs7SoKuxOJdX37k7sBJsPALWEo5ayflizuaH0CAADIAWtjYW5pc3Rlcl9pZEoAAAAAAcDR1wEBbmluZ3Jlc3NfZXhwaXJ5GxeNX/65y4YAa21ldGhvZF9uYW1laHRyYW5zZmVyZW5vbmNlUFF4+hAimFhoqkdUcIchz0xscmVxdWVzdF90eXBlZGNhbGxmc2VuZGVyWB1q63Snu+4C5/fpWFu4nq1IpZxCYDEYA8XSPqPfAg==",
            "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
          }
        },
        {
          "error": {
            "code": 1001,
            "message": "Not processed due to batch request failure"
          }
        }
      ]
    }
  }
}
```

sea-snake:
Maximum number of calls is not fixed and up to the signer which can implement custom rules depending on the canisters and methods called e.g. 20 approvals for low amounts is less concerning than 10 transfers with high amounts.

Todo: Define "too many requests" error.

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,461 @@
+# ICRC-X: Call batch canisters
```
sea-snake, line 1:
```suggestion
# ICRC-X: Batch Call canisters
```

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,461 @@
...
+
+        S ->> U: Display success/failed message
+        S ->> RP: Return the batch call response
+
+        else Rejected
+            U ->> S: Reject request
+            S ->> U: Display reject message
+            S ->> RP: Error response: Action aborted (3001)
+        end
+    end
```
sea-snake, line 179: Updated the sequence diagram to use `par` blocks and do an early check if validation is unavailable while there are multiple sub-arrays in requests.

To simplify the sequence diagram, I removed the details regarding adding responses to a list of responses and returning this whole list, this is already clarified in the examples.

```suggestion
sequenceDiagram
    participant RP as Relying Party
    participant S as Signer
    participant U as User
    participant C as Target Canister
    participant VC as Validate Canister

    RP ->> S: Request batch canister call
    alt Relying party has not been granted the `icrcX_call_canister` permission scope<br>or the request does not comply with scope restrictions
        S ->> RP: Error response: Permission not granted (3000)
    else
        alt Validation unavailable and multiple sub-arrays in requests
            S ->> RP: Error response: Validation required (1002)
        end
        par For all requests
            alt Canister supports ICRC-21
                Note over S,C: Follow the ICRC-21 standard
            else Canister does not support ICRC-21 and signer does not support blind signing
                S ->> RP: Error response: No consent message (2001)
                Note over RP,S: Abort interaction, do not proceed further.
            else Canister does not support ICRC-21 and signer supports blind signing
                S ->> U: Display warning and canister call details (canisterId, sender, method, arg)
                Note over S,U: The warning must inform the user that the canister does not support ICRC-21<br/>The arguments should be decoded, otherwise a stronger warning must be displayed
            end
        end
        alt Approved
            U ->> S: Approve request
            loop For every sub-array in requests
                par For each request
                    S ->> C: Submit canister call
                    S ->> S: Wait for the canister call result
                    alt Call failed
                        S ->> U: Display failure message
                        S ->> RP: Error response: Network error (4000) | Generic error 
                    end
                end
                alt Validation available
                    S ->> VC: Send list with all requests and their responses
                    VC ->> S: Return passed or failed
                    alt Failed
                        S ->> U: Display failed message
                        S ->> RP: Error response: Validation failed (1003)
                    end
                end
            end
            S ->> U: Display success message
            S ->> RP: Return the batch call response
        else Rejected
            U ->> S: Reject request
            S ->> U: Display reject message
            S ->> RP: Error response: Action aborted (3001)
        end
    end
```

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,461 @@
...
+```
+
+### Examples
+
+#### Successful Batch Call
+
+Parallel approve request, swap and then call birdge method
```
sea-snake, line 186:
```suggestion
Parallel approve request, swap and then call bridge method
```

topics/icrc_X_multiple_canister_call.md
```diff
@@ -0,0 +1,461 @@
...
+            "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
+          }
+        }
+      ],
+      [
+        {
+          "error": {
+            "code": 1002,
+            "message": "Validation failed"
+          }
```
sea-snake, line 436: Added content map (request args) and certificate (response) to the validation error, this enables the relying party to also verify the validation call besides the canister calls.

```suggestion
          "error": {
            "code": 1002,
            "message": "Validation failed",
            "contentMap": "2dn3p2NhcmdYTkRJREwEbXtuAGwCs7DawwNorYbKgwUBbAP7ygECot6U6wYB2KOMqA19AQMBHVfs7SoKuxOJdX37k7sBJsPALWEo5ayflizuaH0CAADIAWtjYW5pc3Rlcl9pZEoAAAAAAcDR1wEBbmluZ3Jlc3NfZXhwaXJ5GxeNX/65y4YAa21ldGhvZF9uYW1laHRyYW5zZmVyZW5vbmNlUFF4+hAimFhoqkdUcIchz0xscmVxdWVzdF90eXBlZGNhbGxmc2VuZGVyWB1q63Snu+4C5/fpWFu4nq1IpZxCYDEYA8XSPqPfAg==",
            "certificate": "2dn3omR0cmVlgwGDAYIEWCAPzKZJY/emKhi2GGtBrnHh4cdttATd4+9GtJrNCBepb4MBgwJOcmVxdWVzdF9zdGF0dXODAYIEWCCCgynUaonrKCCywghWCSk9BeDqMoI4yf15nxyU/5JZv4MBggRYIDG7WdzQ9sGWI1MpxizUzxubsEBuNkTT94UOZ9USbzNvgwGCBFggawwbTHxnPUzBAUhWBRjk0nzPs2fPpJlaIYtj5AvcX+ODAYIEWCDiFLyaWuMWjtVurCQcSgny/cqfM8S6qrdihVq7nPz1FoMCWCD/8jdeccvqHVYf06Hw7qPXIDNimC1Uyf47VsvgqKpPiIMBgwJFcmVwbHmCA1RESURMAWsCvIoBfcX+0gFxAQAABIMCRnN0YXR1c4IDR3JlcGxpZWSCBFgg7qZngcNt2+B/RuF44W3LRsKWXG6QQg2L6GdZgJ6Nb3+DAYIEWCAx3tU/mhHfX+wDzF003eSJYN8Nebou8rTeGyxr/rUa1YMCRHRpbWWCA0nw9+r88fjXxhdpc2lnbmF0dXJlWDCXNshvwWG1jGViP7ELePGHCThBw9mts45FxIy4gZATkUEsPeJ6y+cjbn2REmB0Soo="
          }
```

sea-snake (commented):
Added suggestions for spelling and some refactor.

Overall looks good!

baolongt:
@sea-snake since last change we realized we didn't cover all of the requirement by ICRC-112

We're thinking about putting `callback` to call for every single request

so the RPC call should be like this 
```
 "params": {
    "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
    "validation": {
      "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
      "method": "validate_batch_approve_swap_bridge"
    },
     "callback": {
      "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
      "method": "handle_call_back"
    },
    "requests": []
```

We want to make additional canister call for every single canister call made in ICRCX either parallel or sequence  

What do you think?

sea-snake:
> @sea-snake since last change we realized we didn't cover all of the requirement by ICRC-112
> 
> We're thinking about putting `callback` to make additional canister call for every single request either parallel or sequence
> 
> so the RPC call should be like this
> 
> ```json
>  "params": {
>     "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
>     "validation": {
>       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
>       "method": "validate_batch_approve_swap_bridge"
>     },
>      "callback": {
>       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
>       "method": "handle_call_back"
>     },
>     "requests": []
> ```
> 
> What do you think?

Could you go into more detail what a callback would be used for? Since the validation canister is already called for every call.

baolongt:
> > @sea-snake since last change we realized we didn't cover all of the requirement by ICRC-112
> > We're thinking about putting `callback` to make additional canister call for every single request either parallel or sequence
> > so the RPC call should be like this
> > ```json
> >  "params": {
> >     "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
> >     "validation": {
> >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> >       "method": "validate_batch_approve_swap_bridge"
> >     },
> >      "callback": {
> >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> >       "method": "handle_call_back"
> >     },
> >     "requests": []
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think?
> 
> Could you go into more detail what a callback would be used for? Since the validation canister is already called for every call.

The validation is for deciding should the batch call continue or not. But in our case, it shouldn't decide it, we just need callback to update some prop in backend side

sea-snake:
> > > @sea-snake since last change we realized we didn't cover all of the requirement by ICRC-112
> > > We're thinking about putting `callback` to make additional canister call for every single request either parallel or sequence
> > > so the RPC call should be like this
> > > ```json
> > >  "params": {
> > >     "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
> > >     "validation": {
> > >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> > >       "method": "validate_batch_approve_swap_bridge"
> > >     },
> > >      "callback": {
> > >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> > >       "method": "handle_call_back"
> > >     },
> > >     "requests": []
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What do you think?
> > 
> > 
> > Could you go into more detail what a callback would be used for? Since the validation canister is already called for every call.
> 
> The validation is for deciding should the batch call continue or not. But in our case, it shouldn't decide it, we just need callback to update some prop in backend side

But can't you do that already with the validation method? Also keep in mind that you can't trust the candid encoded validation arguments (canister, method, arg and res). Since you're not receiving any certificate with a signature you can trust for example. 

Which is why I asked for a use case example of such a callback, compared to the frontend receiving all the response certificates and making a call with a user identity from your frontend?

baolongt:
> > > > @sea-snake since last change we realized we didn't cover all of the requirement by ICRC-112
> > > > We're thinking about putting `callback` to make additional canister call for every single request either parallel or sequence
> > > > so the RPC call should be like this
> > > > ```json
> > > >  "params": {
> > > >     "sender": "b7gqo-ulk5n-2kpo7-oalt7-p2kyl-o4j5l-kiuwo-eeybr-dab4l-ur6up-pqe",
> > > >     "validation": {
> > > >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> > > >       "method": "validate_batch_approve_swap_bridge"
> > > >     },
> > > >      "callback": {
> > > >       "canisterId": "zzzzz-fqaaa-aaaao-a2hlq-ca",
> > > >       "method": "handle_call_back"
> > > >     },
> > > >     "requests": []
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > What do you think?
> > > 
> > > 
> > > Could you go into more detail what a callback would be used for? Since the validation canister is already called for every call.
> > 
> > 
> > The validation is for deciding should the batch call continue or not. But in our case, it shouldn't decide it, we just need callback to update some prop in backend side
> 
> But can't you do that already with the validation method? Also keep in mind that you can't trust the candid encoded validation arguments (canister, method, arg and res). Since you're not receiving any certificate with a signature you can trust for example.
> 
> And yes, you could send the whole response certificate to the canister, but validating it's BLS signature within a canister seems like a over complication I'd recommend to avoid in any architecture design.
> 
> Which is why I asked for a use case example of such a callback, compared to the frontend receiving all the response certificates and making a call with a user identity from your frontend?

I can show origin flow that we made 

[](https://mermaid.live/edit#pako:eNqlVV1r2zAU_SsXvS4t-XCa1A-F0I5RSPawNiuMQFDl60TUkTxJbraV_vddyXbqxO76MAeCLZ177rlXR9ILEzpBFjOLPwtUAm8k3xi-WymgJ-fGSSFzrhwsLZqO0QfgNszBA88ydG3IdSZRdYzPfeRcqie4Q_MsBbYhs4XH3BuuLBdOagULrvimS8h9oLvXT6hgjknA-B80niDz7OqKVMcwEwJzB8bXbd0x8Kt2CPqZ0MuHHlHHcOcoFzxyJ7bg3vQch_mHZw5uMNdWOnBBzT_h1D8vyKe4pv6BFEYM1iEkpfROA1ph9B72je42H6q7rugbusIoeMy0eFrLpI3FzCLM8txQaZAikthM71Xd-Y90Dde8jtWGRvKywP_UpBL4qPmfCdPR-pPFLRWXbjskFlxJ64hL-CoM2lwre1JvGXIabSRSpW6LPqrI3Hsxn8ilVZOKPOEO11I5P7eXbtuWfdAApwWQ2euWf0HvtYy3lqbC-IzfeSYpnaesoMCpT6UGsI67woJOm7nDd6nuneSeeHnKUEd0L9Js0dwhYsvNJpjr3eIafqp9vk6N3nmzH4xf-K3aYqmM5VU2jQWnxmqr8xZqajvG-017m4bFrhavql5asAUdFNaG5r4x1MOdVc4PTQzHW8l1jGwV0jJZF_PZqb0rtVKl-kNPV5J8kctb8ozFBMg8_rvpT68Wj8_O-p32qn9lPbZDs-MyoXvjxU-tGLHscMVielVYEF-2Yiv1SlBeOH33WwkWO1NgjxldbLYsTjkdRj1W2rW6dA6jmEinzaK8mcIF1WN0yrP4hf1i8Wh0PhlPov7ldHpB_6OLHvsdRgfRaHg56Uf9_ng47EevPfZHa2Ltn19Oo8EwGkxHF-NoMh0PA92PMOlVvf4FbGsZgw)

We're tracking is the `icrc1_transfer` or `icrc2_approve` is success or not and update its status through transaction manager 

The problem is we are not sure the latency of `icrc1_transfer` and `icrc2_approve` it can take 3 minutes for `icrc2_approve` while `icrc1_transfer` is finished. In this case it is not reflect in our system, we have to wait until `icrc2_approve` finished then all state will be updated. 

So we think about a solution can callback after each single request. I thought about the validation canister but it have two things: 
- If validation failed it will stop the execution which is likely not what we want
- It is not clear that when the validation is be triggered, after each request or after the sub-array finished and validate for next sub-array

sea-snake:
> * If validation failed it will stop the execution which is likely not what we want

That's the intend of a batch canister call, the next sequential calls are not made if any of the current parallel calls fail. But all current calls that are in parallel must and will be made simultaneously. The frontend will need to handle the scenario where not all sequence steps were executed.

Looking at your graph, you'd need to validate the responses from the wallet (certificate) to make sure the wallet didn't lie. [By sending the responses to your user (frontend) you can validate the content maps and certificates in this frontend](https://mermaid.live/edit#pako:eNqlVVFr2zAQ_iuHnjaWFqeJk8YPhdKNUUj2sDYtjEBQ5XMi6kieJLfrSv_7TrKd2k1CN6ZAsKTTd9_dfSc9M6FTZAmz-LNEJfCz5CvDNwsFNApunBSy4MrB3KLZs3oL3IY9uOV5jm7X5CKXqPasT_3JqVT3cIXmQQrcNTmfeZtrw5XlwkmtYMYVX-0jch3grvU9KphiGmz8D1oj0Dw6OyPWCZwLgYUD4-O2rmv4TTsE_UDW89seQSdw5cgX3HEn1uBe-XSP1Zxgd5XS5P16pAtKE0hhRH8ZgDLy4jSgFUY_wmMrie1B4TXEv6MrjYK7XIv7pUz_ztnJkheFoZAg04ZWCvJJqfo_R6jS9_L2RaX7svamLhVhXx7vVaAknpxOEpLMnkBwJa0jVOHDMWgLrSxa-BCogUztxwOFrrSXeGIe6FG69S6dLeJhwAqnA3lTcZNZxUpaEHoj1QoyozdQegYffAxuXdpXD2QWgpKYHnTxiXRfl64sUu5wKZXze_9Ev4tOndRo4it6IedcNR33xsY7v-G5JM8evTYN9ajogHXcUVA6a9MI84roAeceeP4WoTmxX0bns3b7iTU3K1Iw4sHgWoJvumsZCkIttm23UJwdlFr5nmVb-SDT99h5kbe5de157uCSUrXGOtgmetKCLekWsjYk9xWhWd4b5XSbxHB3Vlhdy51ASCFl3k7yPuSjrbTrUzVbqTL9bjfUlHyQ80vSjMUUSDx-3paqZ4vdi7n5DpfJQrEe26DZcJnSo_TstxaMUDa4YAl9KiwJL1-whXohU146ffWkBEucKbHHjC5Xa5ZkPLc0q-Rav2jbVXoufmi9aY7QlCXP7BdL-lH_eDQ6jUaDeBifDCfRoMeeWBJHx_3JYDyM6O90MIzilx77HQCi48l4OIpH8XjcP5nEw2jUY5hKp82selTD2_ryB_VoOac). Afterwards you can notify your client (backend) about the status of your canister calls. As seen in your graph, I'd still recommend to verify the icrc1 balance and icrc2 allowance instead of relying on the responses returned by the wallet, even if you forward these through your user (frontend), the user could still send different data. Unless your frontend sends the contentmap and certificate "as is" and validate the signatures in your backend.

> It is not clear that when the validation is be triggered, after each request or after the sub-array finished and validate for next sub-array.

I'd assume validation is called in parallel for each individual parallel call within the sub-array as soon as each individual call has received a response. I can't see a reason to wait with validation e.g. until all sub-array calls have completed. We can specify this in ICRC-112 explicitly.

> The problem is we are not sure the latency of icrc1_transfer and icrc2_approve it can take 3 minutes for icrc2_approve while icrc1_transfer is finished. In this case it is not reflect in our system, we have to wait until icrc2_approve finished then all state will be updated.

Why would your system need to take action before all of the calls have succeeded? Also calls on the IC are either guaranteed (It'll always happen within a given time) or best effort (you'll know within a given window if it failed). So a call that takes 3 minutes isn't something I'd expect.

